PR #9, opened 2023-03-07
Add author pictures to dialects and vocabularies

jviotti:
Signed-off-by: Juan Cruz Viotti <jv@jviotti.com>


jviotti:
Screenshot 2023-03-07 at 16 40 30

PR #19, opened 2023-04-19
Add flex to body and fill the main content

janka102:
Hey I found this website today and think it's a great reference!
I noticed that when the main content doesn't fill the page vertically, then the footer takes more space than needed.

I added display flex to the body and flex filled the main content. This expands the main content to fill the empty space and collapses the footer to minimum height.

Here's a before and after of the change:
**before**:
before
after

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -41,7 +41,7 @@
       }
     </script>
   </head>
-  <body class="h-100 bg-light">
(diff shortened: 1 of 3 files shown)


jviotti:
Hey @janka102 . Thanks for this! Looks like I was not actually watching this repo, so I missed the notification. Apologies for the delay!

PR #21, opened 2023-04-25
Adds integer type

nyoungstudios:
According to the link, https://json-schema.org/draft/2020-12/json-schema-validation.html#section-6.1.1-2, in this markdown document, `"integer"` is also a valid type. 

It says:

> String values MUST be one of the six primitive types ("null", "boolean", "object", "array", "number", or "string"), or "integer" which matches any number with a zero fractional part.


jviotti:
Thanks a lot! Looks like for some reason I was not watching this repo and miss the notification!

PR #43, opened 2024-02-01
Add description and examples to some `core` keywords

AgniveshChaubey:
This PR adds description and examples for `$id`, `$schema`, `$ref` and `$defs` keywords of `core` vocab.

### DIFF
--- a/content/2020-12/core/id.markdown
+++ b/content/2020-12/core/id.markdown
@@ -1,7 +1,7 @@
 ---
 keyword: "$id"
 signature: "URI Reference"
-summary: 'The "$id" keyword declares an identifier for the schema and establishes the base URI for resolving other URI references within the schema. The "$id" keyword is resolved against the base URI of the overall object.'
+summary: 'The `$id` keyword declares an identifier for the schema and establishes the base URI for resolving other URI references within the schema. This keyword is resolved against the base URI of the overall object.'
 kind: ["identifier"]
 instance: ["any"]
 specification: "https://json-schema.org/draft/2020-12/json-schema-core.html#section-8.2.1"
@@ -17,7 +17,7 @@ related:
 
 ## Examples
 
-{{<schema "Declaring an Identifier for the Schema">}}
+{{<schema `Declaring an Identifier for the Schema`>}}
 {
   "$id": "http://example.com/schemas/address.json",
   "type": "object",
@@ -30,7 +30,7 @@ related:
 -  _The __$id__ keyword declares the identifier "http://example.com/schemas/address.json" for the schema. This identifier can be used to reference the schema from other parts of the document or external documents._
 
 
-{{<schema "Establishing Base URI for Resolving References">}}
+{{<schema `Establishing Base URI for Resolving References`>}}
 {
   "$id": "https://example.com/main-schema",
   "type": "object",
--- a/content/2020-12/core/id.markdown
+++ b/content/2020-12/core/id.markdown
@@ -1,9 +1,9 @@
 ---
 keyword: "$id"
 signature: "URI Reference"
-summary: "This keyword declares an identifier for the schema resource."
-kind: [ "identifier" ]
-instance: [ "any" ]
+summary: 'The "$id" keyword declares an identifier for the schema and establishes the base URI for resolving other URI references within the schema. The "$id" keyword is resolved against the base URI of the overall object.'
+kind: ["identifier"]
+instance: ["any"]
 specification: "https://json-schema.org/draft/2020-12/json-schema-core.html#section-8.2.1"
 metaschema: "https://json-schema.org/draft/2020-12/meta/core"
 index: -999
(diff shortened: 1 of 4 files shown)


content/2020-12/core/ref.markdown
```diff
@@ -1,7 +1,7 @@
 ---
 keyword: "$ref"
 signature: "URI Reference"
-summary: "This keyword is used to reference a statically identified schema."
+summary: "The `$ref` keyword is used to reference a statically identified schema. This is useful for avoiding code duplication and promoting modularity when describing complex data structures. The value of `$ref` can be either:  ___Relative___, i.e., using JSON pointers starting with ___\"#/\"___ to reference other components within the same schema document, or ___Absolute___, i.e., a full URI pointing to an external schema definition."
```
jviotti, line 4: > The value of `$ref` can be either:  ___Relative___, i.e., using JSON pointers starting with ___\"#/\"___ to reference other components within the same schema document, or ___Absolute___, i.e., a full URI pointing to an external schema definition.

This is not correct. The correct terminology is:

- `$ref` is set to a URI reference
- A URI reference may be relative or absolute (here probably worth linking to RFC 3986?)
- A URI with a JSON Pointer in a URI fragment (i.e. `#/foo/bar`) is not necessarily relative. These fragments apply to both relative and absolute URIs. For example: `foo#/bar`, `https://example.com#/foo/bar`. Also, if you talk about URI fragment encoding of JSON Pointers, probably worth linking to the JSON Pointer spec, where this is formally defined
- There is no such thing as a "full URI". I guess you meant absolute?
- Note that an absolute URI is not necessarily a remote reference. You can have an absolute URI pointing at a local schema (if the schema declares nested `$id` or if it points to itself) and you can also have a relative URI pointing to a remote one by taking advantage of base URI resolution

Let me know if that makes sense. Happy to elaborate further!

content/2020-12/core/ref.markdown
```diff
@@ -21,3 +21,78 @@ related:
   - vocabulary: core
     keyword: $defs
 ---
+
+## Examples
+
+{{<schema `Using a relative reference:` >}}
+{
+  "$id": "http://example.com/schemas/product.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
jviotti, line 30: Let's stick to 2020-12 for all of these, as we are only documenting that dialect

content/2020-12/core/ref.markdown
```diff
@@ -21,3 +21,78 @@ related:
   - vocabulary: core
     keyword: $defs
 ---
+
```
jviotti, line 24: You should also clarify that the destination of `$ref` should be a schema, but does not need to be. You can find some interesting examples of the latter here: https://github.com/python-jsonschema/referencing-suite

content/2020-12/core/id.markdown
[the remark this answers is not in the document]
jviotti: In particular, nested `$id`s when the values are not URLs is a very interesting case to exemplify

content/2020-12/core/ref.markdown
```diff
@@ -21,3 +21,78 @@ related:
...
+{{</instance-pass>}}
+
+{{<instance-fail `Instance missing the required properties is invalid` >}}
+{
+  "name": "Gadget"
+}
+{{</instance-fail>}}
+
+
+{{<schema `Using an absolute referece:` >}}
```
jviotti, line 66:
```suggestion
{{<schema `Using an absolute referece to the previous schema:` >}}
```

Or something like that to clarify?

content/2020-12/core/ref.markdown
```diff
@@ -21,3 +21,78 @@ related:
...
+{
+  "$id": "http://example.com/schemas/order.json",
+  "type": "object",
+  "properties": {
+    "items": {
+      "type": "array",
+      "items": { "$ref": "http://example.com/schemas/product.json" }
```
jviotti, line 73: Talking about how relative URIs can reference remote schemas, you could exemplify this as follows:

```
"$ref": "schemas/product.json"
```

content/2020-12/core/schema.markdown
```diff
@@ -16,3 +16,17 @@ related:
   - vocabulary: core
     keyword: $defs
 ---
+
+## Examples
+
+{{<schema `This declaration indicates that the schema follows the specifications outlined in JSON Schema Draft 2020-12`>}}
```
jviotti, line 22:
```suggestion
{{<schema `This declaration indicates that the schema is described by the JSON Schema 2020-12 dialect`>}}
```

content/2020-12/core/schema.markdown
```diff
@@ -16,3 +16,17 @@ related:
...
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema"
+}
+{{</schema>}}
+
+{{<schema `This declaration indicates that the schema follows the specifications outlined in JSON Schema Draft 7`>}}
+{
+  "$schema": "http://json-schema.org/draft-07/schema#"
```
jviotti, line 30: This entire section of learnjsonschema.com is purely about the 2020-12 dialect. Instead of showcasing another different dialect, what about exemplifying the HyperSchema variant of 2020-12?

content/2020-12/core/schema.markdown
```diff
@@ -16,3 +16,17 @@ related:
   - vocabulary: core
     keyword: $defs
 ---
+
```
jviotti, line 19: Here you can clarify how a schema is expected to successfully validate against the meta-schema it declares. Maybe show one example where that's not the case and thus it results in an error?

jviotti: We should also talk about what happens if `$schema` is NOT defined.

jviotti: Finally, an interesting case worth explaining and exemplifying is embedded schemas (https://json-schema.org/blog/posts/bundling-json-schema-compound-documents), and how these can declare their own `$schema` as they represent different schema resources

jviotti: Finally, talking about schema resources, let's explain/exemplify that `$schema` must only be declared at the TOP LEVEL of a schema resource.

jviotti (changes requested):
Great start! I added various comments/suggestions. It might be worth splitting this PR and sending one PR for each keyword, that way we can go deep into specific keywords while merging the ones that are done?

Also, `$id`, `$ref`, and `$schema` are quite fundamental and have their own dark corners. I tried to explain various aspects in the review, at least in a high level manner, but happy to go into more details if needed.

PR #45, opened 2024-02-02
add description and examples for `$id` keyword

AgniveshChaubey:
This is a follow-up PR addressing feedback from #43 and making suggested changes.

Not 100% sure how URN works in a schema. Tried with an example, but need some guidance here.

_(Last section will be updated after the review)_

### DIFF
--- a/content/2020-12/core/id.markdown
+++ b/content/2020-12/core/id.markdown
@@ -183,3 +183,27 @@ _**Note:** Check out the [URI RFC](https://datatracker.ietf.org/doc/html/rfc3986
 {{</instance-pass>}}
 
 - A tag URI, (defined in [RFC 4151](http://www.faqs.org/rfcs/rfc4151.html)), is a type of URN used for uniquely identifying resources, typically within a specific context or domain. It consists of a 'tag:' scheme followed by a date and a unique string, providing a human-readable and globally unique identifier. In JSON Schema, a tag URI can be used as the value for the `$id`` keyword to uniquely identify the schema.
+
+ ---
+
+{{<schema `Clarifying schema terminology`>}}
+{
+  "$schema": "https://json-schena.org/draft/2020-12/schema",
+  "$id": "https://example.com",
+  "type": "object",
+  "properties": {
+    "foo": {
+      "$id": "foo",
+      "type": "array",
+      "items": { "type": "boolean" }
+    },
+    "bar": {
+      "$id": "bar",
+      "type": "number"
+    },
+    "baz": { "type": "string" }
+  }
+}
+{{</schema>}}
+
+-  Whenever a schema object has `$id`, a new ***schema resource*** is introduced. In our case, we have three schema resources: one with the `https://example.com` id, one with the `foo` id, and one with the `bar` id. The `https://example.com` schema resource is the ***root schema resource***.
--- a/content/2020-12/core/id.markdown
+++ b/content/2020-12/core/id.markdown
@@ -14,3 +14,121 @@ related:
   - vocabulary: core
     keyword: $vocabulary
 ---
+
+The `$id` keyword declares an identifier for the schema and establishes the base URI for resolving other URI references within the schema. This keyword is resolved against the base URI of the overall object. The top level `$id` of a schema, if present, should be absolute.
+
+## Examples
+
+{{<schema `Declaring an identifier for the schema`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/schemas/address.json",
+  "type": "string"
+}
+{{</schema>}}
+
+{{<instance-pass `A string is valid`>}}
+"Hello World!"
+{{</instance-pass>}}
+
+
+- _The `$id` keyword declares the identifier `http://example.com/schemas/address.json` for the schema. This identifier can be used to reference the schema from other parts of the document or external documents._
+
+{{<schema `Nested subschema with relative $id`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/main-schema",
+  "type": "object",
+  "properties": {
+    "name": { "type": "string" },
+    "age": { "type": "number" },
+    "currentAddress": {
+      "$id": "/address",
+      "type": "object",
+      "properties": {
+        "city": { "type": "string" },
+        "postalCode": { "type": "number" }
+      },
+      "required": ["city", "postalCode"]
+    },
+    "permanentAddress": {
+      "$ref": "https://example.com/address"
+    }
+  }
+}
+{{</schema>}}
+
+{{<instance-pass `An instance with all the required properties is valid`>}}
+{
+  "name": "John Doe",
+  "age": 30,
+  "currentAddress": {
+    "city": "Example City",
+    "postalCode": 12345
+  },
+  "permanentAddress": {
+    "city": "Another City",
+    "postalCode": 67890
+  }
+}
+{{</instance-pass>}}
+
+- The base URI for this schema is `https://example.com/main-schema`. The _address_ subschema has a relative URI `/address`, which resolving against the base URI will result into `https://example.com/address`. Now this URI can be used to reference the  _address_ schema from other parts of the document or external documents.
+
+{{<schema `Nested subschema with absolute $id`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/family-info",
+  "type": "object",
+  "properties": {
+    "name": {
+      "$id": "https://example.com/name",
+      "type": "string"
+    },
+    "fatherName": { "$ref": "https://example.com/name" },
+    "motherName": { "$ref": "https://example.com/name" }
+  },
+  "required": ["name", "fatherName", "motherName"]
+}
+{{</schema>}}
+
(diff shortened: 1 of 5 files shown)


content/2020-12/core/id.markdown
```diff
@@ -14,3 +14,121 @@ related:
...
+  "fatherName": "Peter",
+  "motherName": "Julia"
+}
+{{</instance-pass>}}
+
+- Here, the _name_ subschema has an absolute URI `https://example.com/name`, which can be used to reference the  _name_ schema from other parts of the document or external documents.
+
+{{<schema  `Schema with URN as value of $id`>}}
+{
+  "$id": "urn:example:vehicle",
```
AgniveshChaubey, line 107: I missed adding `$schema` here, I will add it later on

AgniveshChaubey:
I am not sure why the above check is still pending, even though the `website-build` workflow has been run successfully.
image

content/2020-12/core/id.markdown
```diff
@@ -14,3 +14,121 @@ related:
   - vocabulary: core
     keyword: $vocabulary
 ---
+
+The `$id` keyword declares an identifier for the schema and establishes the base URI for resolving other URI references within the schema. This keyword is resolved against the base URI of the overall object. The top level `$id` of a schema, if present, should be absolute.
```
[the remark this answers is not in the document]
jviotti: Now that I think about it, an interesting point we should clarify on a paragraph here is that if the identifier is a URL (i.e. an HTTP one), it is *common* for such URL to respond with the schema if i.e. resolved on a web browser, but that's not mandatory, as the URL is merely an identifier.

content/2020-12/core/id.markdown
```diff
@@ -14,3 +14,121 @@ related:
...
+
+## Examples
+
+{{<schema `Declaring an identifier for the schema`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/schemas/address.json",
```
jviotti, line 25: This says the schema is an address, but the example below is just "Hello World!" Maybe put a fake address in the example instead?

content/2020-12/core/id.markdown
```diff
@@ -14,3 +14,121 @@ related:
...
+  "type": "string"
+}
+{{</schema>}}
+
+{{<instance-pass `A string is valid`>}}
+"Hello World!"
+{{</instance-pass>}}
+
+
+- _The `$id` keyword declares the identifier `http://example.com/schemas/address.json` for the schema. This identifier can be used to reference the schema from other parts of the document or external documents._
```
jviotti, line 35:
```suggestion
- _The `$id` keyword declares the URI `http://example.com/schemas/address.json` as the identifier for the schema. This URI serves as the base URI for resolving other URIs within the schema resource._
```

Or something like that?

content/2020-12/core/id.markdown
```diff
@@ -14,3 +14,121 @@ related:
...
+{{<schema `Nested subschema with relative $id`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/main-schema",
+  "type": "object",
+  "properties": {
+    "name": { "type": "string" },
+    "age": { "type": "number" },
+    "currentAddress": {
+      "$id": "/address",
```
jviotti, line 46:
```suggestion
      "$id": "address",
```

Probably simpler?

content/2020-12/core/id.markdown
```diff
@@ -14,3 +14,121 @@ related:
...
+      "type": "object",
+      "properties": {
+        "city": { "type": "string" },
+        "postalCode": { "type": "number" }
+      },
+      "required": ["city", "postalCode"]
```
jviotti, line 52:
```suggestion
      "required": [ "city", "postalCode" ]
```

Spaces for consistency?

content/2020-12/core/id.markdown
```diff
@@ -14,3 +14,121 @@ related:
...
+    },
+    "permanentAddress": {
+      "$ref": "https://example.com/address"
```
jviotti, line 55:
```suggestion
      "$ref": "address"
```

Because the top level identifier is already based on `https://example.com`.

content/2020-12/core/id.markdown
```diff
@@ -14,3 +14,121 @@ related:
...
+{
+  "name": "John",
+  "fatherName": "Peter",
+  "motherName": "Julia"
+}
+{{</instance-pass>}}
+
+- Here, the _name_ subschema has an absolute URI `https://example.com/name`, which can be used to reference the  _name_ schema from other parts of the document or external documents.
+
+{{<schema  `Schema with URN as value of $id`>}}
```
jviotti, line 105:
```suggestion
{{<schema `Schema with URN as value of $id`>}}
```

content/2020-12/core/id.markdown
```diff
@@ -14,3 +14,121 @@ related:
...
+      "$id": "urn:example:vehicle:car",
+      "type": "object",
+      "properties": {
+        "brand": {"type": "string"},
+        "price": {"type": "number" }
+      }
+    }
+  }
+}
+{{</schema>}}
```
jviotti, line 125: The example looks good. Maybe clarify that with URNs there is no such thing as a relative URN. i.e. you always have to fully type them down?

jviotti: Also, can you add one example using Tag URIs (http://www.taguri.org), as JSON Schema is trying to promote them a bit? Maybe add a short paragraph explaining what they are? You can find examples on the referencing suite: https://github.com/python-jsonschema/referencing-suite/blob/main/tests/json-schema-draft-2020-12/tag-uris.json

jviotti: The RFC is also a good read: http://www.faqs.org/rfcs/rfc4151.html

jviotti (changes requested):
Taking really good shape! The URN example is correct

jviotti:
> I am not sure why the below check is still pending, even though the website-build workflow has been run successfully in my fork.

I think GitHub will not report it back until you are either a collaborator or have merged at least one PR. Let me actually make you a collaborator. Then you can also send PRs straight away without requiring a fork

jviotti:
Done!

Screenshot 2024-02-02 at 20 11 38

jviotti:
I'm actually writing a blog post touching on schema resources and `$id`, so you might find these in-progress parts helpful:

Screenshot 2024-02-02 at 20 12 45
Screenshot 2024-02-02 at 20 12 56

jviotti:
`$id`, URIs, and schema resources are definitely confusing 😅 I think it will be great to have another JSON Schema contributor (you!) with a good grasp on them!

content/2020-12/core/id.markdown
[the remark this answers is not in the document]
jviotti: On the last point, we could also say that if you are not planning to make your URIs locatable (i.e. actually accessible over the protocol they declare, like HTTP), we advise to use URNs?

content/2020-12/core/id.markdown
jviotti: (Tag URIs are a *type* of URNs, so might be worth clarifying that)

content/2020-12/core/id.markdown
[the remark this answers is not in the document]
jviotti: Actually, given that URIs are so fundamental to all of this, maybe add a note suggesting the reader to go read the URI RFC to gain a deeper understanding on how resolution works?

content/2020-12/core/id.markdown
AgniveshChaubey: I intentionally kept the absolute URI to explicitly indicate where the `$ref` is being resolved.

AgniveshChaubey:
Addressed all the requested changes.

Thanks for sharing your blog for the reference. However, the last section of your blog seems a bit confusing to me. I mean why _children schema resource_ is not considered as a part of _parent schema resource_? Is there any major difference between schema and schema resource?

Or is the above thing only applicable for the child schemas having the `$id` keyword?

content/2020-12/core/id.markdown
jviotti: Alright!

jviotti:
> However, the last section of your blog seems a bit confusing to me. I mean why children schema resource is not considered as a part of parent schema resource? Is there any major difference between schema and schema resource?

Yeah, this is a common source of confusion that we should describe/exemplify well for this keyword. Essentially:

- A *schema* is the entire JSON boolean or JSON object you are passing to i.e. a validator
- A *schema* may consist of one or more *schema resources*. A schema resource is essentially an `$id` boundary. If you create nested schemas objects with `$id` in your schema, you are introducing new schema resources
- A *schema object* is a single subschema in the schema tree, only considering its immediate keywords and not considering nested subschemas

So the relationship is:

- A *schema* has one or more *schema resources*
- A *schema resource* has one or more *schema objects*
- A *schema object* has one or more *keywords*

Coming back to your original question, a schema resource does not include its children schema resources, as they are conceptually completely different entities, just that they are nested. That said, they are all part of the same *schema*

AgniveshChaubey:
Changes addressed in #106

AgniveshChaubey:
Closing this PR as it is not passing the required checks. Requested changes are addressed in #106

PR #106, opened 2024-02-06
add description and examples for `$id` keyword (#105)

AgniveshChaubey:
This PR addresses #47.

It is built upon the changes requested in #45 and implements the suggested modifications

Also let me know what you think about this card.
image

### DIFF
--- a/assets/main.scss
+++ b/assets/main.scss
@@ -125,3 +125,10 @@ article table {
 .code-instance + .code-schema {
   @extend .mt-4;
 }
+
+.alert {
+  background-color: #fffdbd;
+  padding: 10px;
+  border: 1px solid #ccc;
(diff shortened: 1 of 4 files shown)


jviotti:
BTW, if you are up for it, you should consider upstreaming the definitions of schemas, schema resources, and schema objects to the JSON Schema official glossary cc @Julian

jviotti:
Later, if they are the glossary, we can probably explain them less here and link there instead?

AgniveshChaubey:
> BTW, if you are up for it, you should consider upstreaming the definitions of schemas, schema resources, and schema objects to the JSON Schema official glossary cc @Julian

Sure, I'm happy to help with that. Thanks for the suggestion!

Issue #87, opened 2024-02-05
Document `const`


jviotti:

AgniveshChaubey:
This task was addressed in #107.

PR #107, opened 2024-02-07
add description and examples for `const` keyword

AgniveshChaubey:
This PR addresses #87.

### DIFF
--- a/content/2020-12/validation/const.markdown
+++ b/content/2020-12/validation/const.markdown
@@ -25,7 +25,6 @@ The `const` keyword in JSON Schema restricts a property value to a single, speci
 {{<schema `Schema with a specific string value`>}}
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "type": "string",
   "const": "hello"
 }
 {{</schema>}}
@@ -41,7 +40,6 @@ The `const` keyword in JSON Schema restricts a property value to a single, speci
 {{<schema `Schema with a specific number value`>}}
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "type": "number",
   "const": 3.14159
 }
 {{</schema>}}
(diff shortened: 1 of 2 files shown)


content/2020-12/validation/const.markdown
```diff
@@ -14,3 +14,64 @@ related:
...
+
+{{<instance-fail `An instance not matching the const value is invalid.`>}}
+"pi"
+{{</instance-fail>}}
+
+{{<schema `Schema with a fixed object structure`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "properties": {
```
jviotti, line 59: Similar to `enum`, if using `const`, you should not be declaring any other validation properties. In this case, we should get rid of both `type` and `properties`, as `const` will take precedence over them.

Maybe you can even explain this best practice like you did with `enum`?

content/2020-12/validation/const.markdown
```diff
@@ -14,3 +14,64 @@ related:
   - vocabulary: validation
     keyword: type
 ---
+
+The `const` keyword in JSON Schema restricts a property value to a single, specific constant, applicable to any data type, including null. Its usage is functionally similar to an `enum` with a single value. Instances validate successfully only if their property value precisely matches the specified constant.
```
jviotti, line 18:
```suggestion
The `const` keyword in restricts an instance to a specific value. Its usage is functionally similar to an `enum` with a single value. Instances validate successfully only if their property value deeply matches the specified constant.
```

Saying "property value" sounds confusing, as it's not just about objects? Also adjusted grammar a bit

jviotti (changes requested):
Looks good. Just left some minor comments

AgniveshChaubey:
Done!

jviotti (approved):

Issue #90, opened 2024-02-05
Document `maxLength`


jviotti:

AgniveshChaubey:
This task was addressed in #110 and #112

PR #110, opened 2024-02-12
add description and examples for `maxLength` keyword

AgniveshChaubey:
This PR addresses #90


content/2020-12/validation/maxLength.markdown
```diff
@@ -16,3 +16,49 @@ related:
   - vocabulary: format-annotation
     keyword: format
 ---
+
+The `maxLength` keyword is used to specify the maximum length of a string instance. It is used to enforce a constraint on the maximum number of characters allowed for a string instance.
+
+* Applies only to string data types.
+* Value must be a non-negative integer.
+* String length is counted in characters, not bytes.
```
jviotti, line 24: Now that you mention it (which is a good point), I think it'd be good exemplifying this. I guess you can provide an instance that consists of unicode characters in the UTF-16 and/or UTF-32 range, showing that it still validates against `maxLength` based on the number of characters.

I recall seeing some examples of this in the JSON Schema Test Suite

jviotti (approved):
I left a suggestion, but the rest looks good. I guess we can merge this one and do the other separately if you want

Issue #53, opened 2024-02-05
Document `$anchor`


jviotti:

jviotti:
- Maybe explain how this is analogous to `id` in HTML? I guess most people understand that
- Explain that two equal anchors on the same lexical scope is invalid
- Explain that you can reference an anchor with/without a relative/absolute URI too
- Explain that anchors work with URNs too (I guess? Maybe double check 😅)

AgniveshChaubey:
This task was addressed in #113.

PR #113, opened 2024-02-13
add description and examples for `$anchor` keyword

AgniveshChaubey:
This PR addresses #53.

### DIFF
--- a/content/2020-12/core/anchor.markdown
+++ b/content/2020-12/core/anchor.markdown
@@ -17,3 +17,95 @@ related:
   - vocabulary: core
     keyword: $dynamicAnchor
 ---
+
+
+The `$anchor` keyword is used to assign a unique identifier to a subschema within the scope of its containing schema. This identifier can then be referenced elsewhere in the schema using the `$ref` keyword.
+
+* Its value must be a valid identifier starting with a letter and containing letters, digits, hyphens, underscores, colons, or periods.
+* This keyword allows for the creation of plain name fragments, providing a flexible alternative to using JSON Pointer fragments which require knowledge of the schema's structure.
+* By using `$anchor`, schemas can be written with reusable components that aren't tied to specific structural locations.
+* These subschemas with named anchor are derived from the nearest `$id` or document base URI.
+
+## Examples
+
+{{<schema `Schema with a named anchor (identifier)`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$ref": "#string",
+  "$defs": {
+    "string": {
+      "$anchor": "string",
+      "type": "string"
+    }
+  }
+}
+{{</schema>}}
+
+{{<instance-pass `An instance with a string is valid`>}}
+"Hello World!"
+{{</instance-pass>}}
+
+{{<instance-fail `An instance with a number is invalid`>}}
+44
+{{</instance-fail>}}
+
+{{<schema `Schema with identifiers having absolute URI`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "properties": {
+    "name": { "$ref": "https://example.com/person/name#name" },
+    "age": { "$ref": "https://example.com/person/age#age" }
+  },
+  "required": [ "name", "age" ],
+  "$defs": {
+    "name": {
+      "$id": "https://example.com/person/name",
+      "$anchor": "name",
+      "type": "string"
+    },
+    "age": {
+      "$id": "https://example.com/person/age",
+      "$anchor": "age",
+      "type": "integer"
+    }
+  }
+}
+{{</schema>}}
+
+{{<instance-pass `An instance adhering to the schema is valid`>}}
+{
+  "name": "John",
+  "age": 55
+}
+{{</instance-pass>}}
+
+{{<instance-fail `The value of age must be an integer`>}}
+{
+  "name": "foo",
+  "age": "bar"
+}
+{{</instance-fail>}}
+
+{{<schema `Schema with location-independent identifier having base URI change in subschema`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/base",
+  "$ref": "https://example.com/nested#foo",
+  "$defs": {
+    "foo": {
+      "$id": "nested",
+      "$anchor": "foo",
+      "type": "integer"
+    }
+  }
+}
+{{</schema>}}
+
+{{<instance-pass `An instance with integer is valid`>}}
+99
+{{</instance-pass>}}
+
+{{<instance-fail `An instance with boolean is invalid`>}}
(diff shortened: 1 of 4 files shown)


content/2020-12/core/anchor.markdown
```diff
@@ -17,3 +17,95 @@ related:
   - vocabulary: core
     keyword: $dynamicAnchor
 ---
+
+
```
jviotti, line 21: Extra blank line here?

content/2020-12/core/anchor.markdown
```diff
@@ -17,3 +17,95 @@ related:
...
+The `$anchor` keyword is used to assign a unique identifier to a subschema within the scope of its containing schema. This identifier can then be referenced elsewhere in the schema using the `$ref` keyword.
```
jviotti, line 22:
```suggestion
The `$anchor` keyword is used to assign a unique identifier to a subschema within its schema resource. This identifier can then be referenced elsewhere using the `$ref` keyword.
```

Small clarifications on schema resources, and the fact that you can reference anchors even from outside the current schema

content/2020-12/core/anchor.markdown
```diff
@@ -17,3 +17,95 @@ related:
...
+
+* Its value must be a valid identifier starting with a letter and containing letters, digits, hyphens, underscores, colons, or periods.
```
jviotti, line 24: Is this actually the case? I guess any valid URI fragment would do, but I might be wrong. i.e. `1234` or even something that looks like a JSON Pointer might in the theory do.

Can you cross check this with a few implementations and with the URI RFC?

content/2020-12/core/anchor.markdown
```diff
@@ -17,3 +17,95 @@ related:
...
+* This keyword allows for the creation of plain name fragments, providing a flexible alternative to using JSON Pointer fragments which require knowledge of the schema's structure.
+* By using `$anchor`, schemas can be written with reusable components that aren't tied to specific structural locations.
```
jviotti, line 26: This point seems similar to the previous one. Maybe merge them into one and simplify them?

content/2020-12/core/anchor.markdown
```diff
@@ -17,3 +17,95 @@ related:
...
+* These subschemas with named anchor are derived from the nearest `$id` or document base URI.
```
jviotti, line 27: Maybe a more correct way of saying it would be:

```suggestion
* An anchor is resolved against the base URI of its schema resource
```

content/2020-12/core/anchor.markdown
jviotti, line 1: I think another point worth mentioning is that defining two equal anchors on the SAME schema resource is considered to be an error.

However, defining two equal anchors on DIFFERENT schema resources (i.e. subschemas with their own `$id`) is valid. Maybe show an example of that?

jviotti (changes requested):
Good take! I left some comments but indeed grasp the idea of anchors very well!

content/2020-12/core/anchor.markdown
AgniveshChaubey: Okay, let me verify

AgniveshChaubey: image
https://json-schema.org/draft/2019-09/json-schema-core#rfc.section.8.2.3

I tried with [hyperhump schema validator](https://json-schema.hyperjump.io/) which allows any string values (including special characters) without checking any format (which I think is not an expected behavior), whereas [json everything](https://json-everything.net/json-schema) strictly follows the above rule and only allows the `$achors` with  above-mentioned format.

content/2020-12/core/anchor.markdown
AgniveshChaubey: You mean something like this? ---
Valid:
```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://example.com/base",
  "properties": {
    "foo": {
      "$id": "one",
      "$anchor": "bar",
      "type": "integer"
    },
    "bar": {
      "$id": "two",
      "$anchor": "bar",
      "type": "string"
    }
  }
}
```

Invalid: 
```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://example.com/base",
  "properties": {
    "foo": {
      "$anchor": "bar",
      "type": "integer"
    },
    "bar": {
      "$anchor": "bar",
      "type": "string"
    }
  }
}
```

However, the implementations do not throw any error for the second schema and the fragmented URI reference is resolved to the last `$anchor`

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://example.com/base",
  "properties": {
    "foo": {
      "$anchor": "bar",
      "type": "integer"
    },
    "bar": {
      "$anchor": "bar",
      "type": "string"
    },
    "baz": {
      "$ref": "#bar"
    }
  }
```
Instance: `{"foo": 22, "bar": "aaaa", "baz": "aaaa" }`

AgniveshChaubey:
Apologies for the delayed response. I've been dealing with some family issues, which is making me a bit distracted.

jviotti:
No hurries at all! Take your time :)

content/2020-12/core/anchor.markdown
jviotti: > However, the implementations do not throw any error for the second schema and the fragmented URI reference is resolved to the last $anchor

That's strange. Which implementations are you testing with?

jviotti: @jdesrosiers Shouldn't Hyperjump fail for the latter schema? i.e. same anchor twice in the same schema resource?

content/2020-12/core/anchor.markdown
jviotti: Ah, great

jviotti (approved):
Approving it already, and we can sort out the case with conflicting anchors separately?

content/2020-12/core/anchor.markdown
jdesrosiers: Having duplicate anchors doesn't make the schema invalid. It's just nonsense to do so and the behavior is undefined. It's like having a JSON value with duplicate properties (`{ "a": 1, "a": 2 }`). It's not invalid JSON, but you can't count on different implementations handling it the same way.

So, definitely tell users not to do that, but don't tell them they should expect an error.

content/2020-12/core/anchor.markdown
jdesrosiers: If I remember correctly (and if it matters) this constraint didn't always exist. I think it was added in 2019-09. Sorry, for not looking it up myself.

content/2020-12/core/anchor.markdown
jviotti: Ah, interesting. I thought it would result in an error. Then probably fine to ignore all this @AgniveshChaubey

AgniveshChaubey:
Sure! 👍

Issue #49, opened 2024-02-05
Document `$ref`


jviotti:

jviotti:
This one is probably quite simple:

- Show it off referring to an absolute URI within/outside the schema
- Show it off referring to a relative URI within/outside the schema
- Show it off referring to a URI (relative/absolute) + JSON Pointer within/outside the schema
- Show it off referring to a URI (relative/absolute) + anchor within/outside the schema
- Maybe examples of JUST a pointer and JUST an anchor too?

jviotti:
Ah, you can also show off `$ref` with URNs?

AgniveshChaubey:
This task was addresed in #119.

PR #119, opened 2024-02-28
Keyword `ref`

AgniveshChaubey:
This PR addresses #49.

Changes requested in #43 have also been addressed.

### DIFF
--- a/content/2020-12/core/ref.markdown
+++ b/content/2020-12/core/ref.markdown
@@ -21,3 +21,76 @@ related:
   - vocabulary: core
     keyword: $defs
 ---
+
+The `$ref` keyword is used to statically reference a schema. This is useful for avoiding code duplication and promoting modularity when describing complex data structures.
+* The value of `$ref` is set to a URI reference, which may be either relative or absolute according to [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986).
+* A URI reference may include a [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) in a URI fragment (e.g., `#/foo/bar`), but it's important to note that these fragments apply to both relative and absolute URIs.
+
+{{<alert>}}
+ _**Note:** It's crucial to understand that an absolute URI does not necessarily denote a remote reference. An absolute URI can point to a local schema if the schema declares nested `$id` or if it points to itself. Conversely, a relative URI can point to a remote schema by leveraging base URI resolution._
+{{</alert>}}
+
+## Examples
+
+{{<schema `Using a relative reference:` >}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "http://example.com/schemas/product.json",
+  "type": "object",
+  "required": [ "productId", "name" ],
+  "properties": {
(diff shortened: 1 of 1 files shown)


content/2020-12/core/ref.markdown
```diff
@@ -21,3 +21,196 @@ related:
   - vocabulary: core
     keyword: $defs
 ---
+
+The `$ref` keyword is used to statically reference a schema. This is useful for avoiding code duplication and promoting modularity when describing complex data structures.
+* The value of `$ref` is set to a URI reference, which may be either relative or absolute according to [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986).
+* A URI reference may include a [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) in a URI fragment (e.g., `#/foo/bar`), but it's important to note that these fragments apply to both relative and absolute URIs.
```
jviotti, line 27:
```suggestion
* A URI reference may include a [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) in a URI fragment (e.g., `#/foo/bar`)
```

This feels redundant, given you already said a URI reference may be absolute or relative

content/2020-12/core/ref.markdown
```diff
@@ -21,3 +21,196 @@ related:
...
+
+{{<alert>}}
+ _**Note:** It's crucial to understand that an absolute URI does not necessarily denote a remote reference. An absolute URI can point to a local schema if the schema declares nested `$id` or if it points to itself. Conversely, a relative URI can point to a remote schema by leveraging base URI resolution._
```
jviotti, line 30:
```suggestion
 _**Note:** It's crucial to understand that an absolute URI does not necessarily denote a remote reference. An absolute URI can point to a local schema if the schema declares nested `$id`s or if it points to itself. Conversely, a relative URI can point to a remote schema by leveraging base URI resolution._
```

content/2020-12/core/ref.markdown
```diff
@@ -21,3 +21,196 @@ related:
...
+{{<instance-pass `Instance including all the required properties is valid` >}}
+{
+  "productId": 123,
+  "name": "Widget"
+}
+{{</instance-pass>}}
+
+{{<instance-fail `Instance missing the required properties is invalid` >}}
+{
+  "name": "Gadget"
```
jviotti, line 63: I'm not sure I find this example great for this case. Maybe show off `name` being not a string?

content/2020-12/core/ref.markdown
```diff
@@ -21,3 +21,196 @@ related:
...
+}
+{{</instance-fail>}}
+
+{{<schema `Schema with an absolute referece to the previous schema` >}}
```
jviotti, line 67:
```suggestion
{{<schema `Schema with an absolute reference to the previous schema` >}}
```

jviotti (approved):
I left a few minor comments, but pretty cool examples :+1:

content/2020-12/core/ref.markdown
AgniveshChaubey: Sure!

Issue #96, opened 2024-02-05
Document `dependentRequired`


jviotti:

jviotti:
For this one, I guess we can show a complex schema where one keyword requires another keyword, which might require multiple other keywords? We can also showcase some of these keywords being defined with `properties`/`patternProperties`/etc but some not defined at all?

AgniveshChaubey:
This task was addressed in #120.

PR #120, opened 2024-02-29
keyword `dependentRequired`

AgniveshChaubey:
This PR addresses #96

I wasn't sure how to utilize the patternProperties in a real-world example, so I hardcoded its value. Let me know if you have any suggestions regarding this.

### DIFF
--- a/content/2020-12/validation/dependentRequired.markdown
+++ b/content/2020-12/validation/dependentRequired.markdown
@@ -68,3 +68,54 @@ The `dependentRequired` keyword specifies a conditional dependency between prope
 {{<instance-pass `An empty object is also valid`>}}
 {}
 {{</instance-pass>}}
+
+{{<schema `Complex schema with the 'dependentRequired' keyword `>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "properties": {
+    "productName": { "type": "string" },
+    "productPriceUSD": { "type": "number" },
+    "units": { "type": "number" }
+  },
+  "patternProperties": {
+    "^paymentStatus$": { "enum": [ "success", "pending", "failed" ] }
+  },
+  "dependentRequired": {
+    "productPriceUSD": [ "productName" ],
+    "totalCost" : [ "productPriceUSD", "units" ],
+    "trackingId": [ "outForDelivery", "paymentStatus" ]
+  }
+}
+{{</schema>}}
+
+{{<instance-pass `An instance with all the dependent properties is valid`>}}
+{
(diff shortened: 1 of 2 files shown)


content/2020-12/validation/dependentRequired.markdown
```diff
@@ -20,3 +20,102 @@ related:
   - vocabulary: applicator
     keyword: else
 ---
+
+The `dependentRequired` keyword specifies a conditional dependency between properties within an instance. It ensures that if a certain property is present in an instance, then another specified set of properties must also be present. Essentially, it extends the functionality of required by allowing dependencies based on the existence of other properties. In short, if property A exists in an instance, then properties B, C, and D must also be present.
```
jviotti, line 24:
```suggestion
The `dependentRequired` keyword specifies a conditional dependency between properties within an instance. It ensures that if a certain property is present in an instance, then another specified set of properties must also be present. In short, if property A exists in an instance, then properties B, C, and D must also be present.
```

This sentence seems to just repeat the same thing?

content/2020-12/validation/dependentRequired.markdown
```diff
@@ -20,3 +20,102 @@ related:
...
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "properties": {
+    "productName": { "type": "string" },
+    "productPriceUSD": { "type": "number" },
+    "units": { "type": "number" }
+  },
+  "patternProperties": {
+    "^paymentStatus$": { "enum": [ "success", "pending", "failed" ] }
```
jviotti, line 82:
```suggestion
    "[Ss]tatus": { "enum": [ "success", "pending", "failed" ] }
```

Maybe something like this? (please check my regex to make sure I got it right. That would match every property that has `Status` or `status` on its name, which might be an interesting example

AgniveshChaubey: Yes, the regex is correct, but how would I include that exact property name in the `teackingId` array? `"trackingId": [ "outForDelivery", "Status" ]`. I guess it's not possible, and I have to use a hard-coded property name in that array! isn't it?

jviotti: You can do it with `dependentSchemas`, but indeed not with `dependentRequired`. Maybe let's not exemplify `patternProperties` here, and do so in `dependentSchemas` instead? Otherwise a `patternProperties` that matches a fixed string is not very recommended (you should be using just `properties` instead)

AgniveshChaubey: Updated

Issue #92, opened 2024-02-05
Document `multipleOf`


jviotti:

jviotti:
For this keyword, we can show a positive integer and a positive real number

AgniveshChaubey:
This task was addressed in #121

PR #121, opened 2024-03-02
Keyword `multipleOf`

AgniveshChaubey:
This PR addresses #92

### DIFF
--- a/content/2020-12/validation/multipleOf.markdown
+++ b/content/2020-12/validation/multipleOf.markdown
@@ -22,13 +22,13 @@ related:
 The `multipleOf` keyword is used to specify that an instance must be a multiple of a given number. The value of this keyword must be strictly greater than zero.
 * Applicable only to `number` and `integer` type.
 * Validates if an instance is divisible by the specified number.
+* Setting `multipleOf` to 0 is not valid.
 
 ## Examples
 
 {{<schema `Schema with 'multipleOf' set to an integer`>}}
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "type": "integer",
   "multipleOf": 5
 }
 {{</schema>}}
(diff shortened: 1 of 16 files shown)


content/2020-12/validation/multipleOf.markdown
```diff
@@ -18,3 +18,58 @@ related:
...
+
+{{<instance-fail `2 is not a multiple of 4.1`>}}
+2
+{{</instance-fail>}}
+
+{{<instance-pass `0 is a multiple of 4.1`>}}
+0
+{{</instance-pass>}}
+
+{{<schema `Invalid Schema`>}}
```
jviotti, line 68: I think as we did in another keyword, let's avoid schemas that are actually invalid. I agree it is a good point, but I'm afraid that if we show it, some people will not read the notes and will blindly copy-paste assuming it's valid. Maybe let's remove this and have a note at the top explicitly saying that setting `multipleOf` to 0 is not valid?

content/2020-12/validation/multipleOf.markdown
```diff
@@ -18,3 +18,58 @@ related:
   - vocabulary: validation
     keyword: minimum
 ---
+
+The `multipleOf` keyword is used to specify that an instance must be a multiple of a given number. The value of this keyword must be strictly greater than zero.
+* Applicable only to `number` and `integer` type.
```
jviotti, line 23: It might be worth adding one instance example to each of the schemas below where the instance is not a number. like "foo" being valid on both?

jviotti (approved):
I left a couple comments but looks good overall

content/2020-12/validation/multipleOf.markdown
AgniveshChaubey: "foo" will be invalid in both the cases because of the `type` constraint. However, I have tweaked the examples to demonstrate the point you mentioned above

Issue #99, opened 2024-02-05
Document `required`


jviotti:

jviotti:
We can also show this keyword with/without `properties`, and clarify that the list is expected to not contain any duplicates (in fact validation should fail if so)

AgniveshChaubey:
This task was addressed in #122.

PR #122, opened 2024-03-02
Keyword `required`

AgniveshChaubey:
This PR addresses #99

### DIFF
--- a/content/2020-12/validation/required.markdown
+++ b/content/2020-12/validation/required.markdown
@@ -17,19 +17,77 @@ related:
     keyword: minProperties
 ---
 
-The `required` keyword is used to specify which properties must be present within an object instance.
+The `required` keyword is used in conjunction with the `properties` keyword to specify which properties must be present within an object instance.
 * The value of this keyword must be an array.
 * Elements of this array, if any, must be strings, and must be unique.
 * Omitting this keyword has the same behavior as an empty array.
+* The `required` keyword can also be used within nested object schemas.
 
 ## Examples
 
-{{<schema>}}
-
+{{<schema `Schema with the 'required' keyword`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "properties": {
+    "name": { "type": "string" },
+    "age": { "type": "number" }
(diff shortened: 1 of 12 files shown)


content/2020-12/validation/required.markdown
```diff
@@ -16,3 +16,115 @@ related:
   - vocabulary: validation
     keyword: minProperties
 ---
+
+The `required` keyword is used to specify which properties must be present within an object instance.
+* The value of this keyword must be an array.
+* Elements of this array, if any, must be strings, and must be unique.
+* Omitting this keyword has the same behavior as an empty array.
+* The `required` keyword can also be used within nested object schemas.
```
jviotti, line 24: I think this sentence is a bit confusing. Do we really need to mention it? I guess every keyword we are documenting works in subschemas?

content/2020-12/validation/required.markdown
```diff
@@ -16,3 +16,115 @@ related:
...
+{{<schema>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "properties": {
+    "name": { "type": "string" }
+  },
+  "required": [ "name", "age", "name" ]
+}
+// Schema with duplicate items in the required list is invalid.
```
jviotti, line 129: I didn't check it on the rendered version, but would this comment look good? Maybe add it as a paragraph after the schema element as for other explanations you did?

jviotti (approved):
I left two minor comments, but looks good overall

content/2020-12/validation/required.markdown
AgniveshChaubey: In the examples where I included the sentences as a paragraph, I first provided some general notes and then referenced the above/below example accordingly.

But here, this sentence is specifically targeting the schema above, that's why I have included it as a comment within the schema element rather than as a paragraph.
image

Please let me know if it doesn't look good.

content/2020-12/validation/required.markdown
AgniveshChaubey: Removed!

jviotti (approved):

Issue #104, opened 2024-02-05
Document `uniqueItems`


jviotti:

jviotti:
I guess we can show off this keyword as both `true` and `false` (the default?). We can showcase uniqueness of arrays of scalars (like integers) but also uniqueness on arrays of complex objects too

AgniveshChaubey:
This task was addressed in #127

PR #127, opened 2024-03-11
Keyword `uniqueItems`

AgniveshChaubey:
This PR addresses #104

### DIFF
--- a/content/2020-12/validation/uniqueItems.markdown
+++ b/content/2020-12/validation/uniqueItems.markdown
@@ -40,6 +40,10 @@ The `uniqueItems` keyword is used to ensure that all the items in an array are u
 [ false, "world", 2, 2 ]
 {{</instance-fail>}}
 
+{{<instance-fail `An instance with duplicate complex structures (objects) is invalid`>}}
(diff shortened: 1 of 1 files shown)


content/2020-12/validation/uniqueItems.markdown
```diff
@@ -15,3 +15,80 @@ related:
...
     keyword: contains
 ---
+
+The `uniqueItems` keyword is used to ensure that all the items in an array are unique. This keyword is particularly useful when you need to enforce that an array contains no duplicate elements.
+* The value of this keyword must be a boolean.
+* This keyword, when set to true, specifies that all elements in an array must be unique.
+* If it is set to false, the array can contain duplicate items.
+* Omitting this keyword has the same behavior as a value of false.
+
+## Examples
```
jviotti, line 25: Can we clarify in an example that element uniqueness also deeply applies for complex structures like objects? i.e. maybe extend the first schema example with an instance that has two duplicated complex objects?

jviotti (approved):
Small comment, but good already

Issue #83, opened 2024-02-05
Document `unevaluatedProperties`


jviotti:

jviotti:
- Explain what the concept of "evaluation" means in this case and what does it mean for a property to not be evaluated
- Show it off alongside `properties`, `patternProperties`, etc. Mainly, cover how this keyword is defined based on the *annotations* that these keywords produce
- Explain the annotations this keyword produces 
- Show it being a schema and a boolean (both true and false)
- Maybe show its relationship when the corresponding keywords are not directly siblings to it, but within an `allOf` or through a `$ref`?

jviotti:
This one is an important one to nail down well, as 99% of the community out there is VERY confused about this vocabulary

AgniveshChaubey:
This task was addressed in #128.

PR #128, opened 2024-03-14
Keyword `unevaluatedProperties`

AgniveshChaubey:
This PR addresses #83.

Some of the sentences I have added are based on my interpretation of the results from hyperjump and json-everything. I have tried my best to cover most of the fundamental aspects. If any sentence needs improvement or seems incorrect, please provide a brief explanation why.

Additionally, I haven't removed the annotations section (line 26-29). Let me know if we need to keep it or not.


jviotti:
Awesome! This is a tricky one. Let me review with more time tomorrow!

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,182 @@ related:
 Annotations
 -----------
 
-If this keyword is applied to any instance element, it produces an annotation
-value of `true`.
+If this keyword is applied to any instance element, it produces an annotation value of `true`.
+
+## Explanation
+
+First of all, let us try to understand what _evaluation_ means. `unevaluatedProperties` considers annotations from `properties`, `patternProperties`, and `additionalProperties`, both as adjacent keywords and in subschemas of adjacent keywords. Each of these keywords will produce an annotation of the properties that they've evaluated. For `properties`, it's just the listed keys intersected with what's in the instance; for `patternProperties`, it's the list of properties that matches at least one of the regexes; for `additionalProperties`, it's basically the rest of the properties. If any of these are in subschemas of adjacent keywords, and those subschemas fail validation, those annotations are dropped in that case. The effect is that those properties are not considered evaluated.
```
jviotti, line 33: I think this is on the right path. What I propose, to make things simpler, is:

- Pause this PR for a bit. We can let it open
- Document `properties`, `patternProperties`, and `additionalProperties`. In there, include direct examples of the annotations they produce. i.e. using real examples from JSON Schema standard output formats
- Then, you will be able to simplify the documentation for this keyword a lot by just saying it applies to the keywords not covered by the annotations of those

jviotti, line 33: Keep in mind `unevaluatedProperties` may also be affected by other `unevaluatedProperties` in nested schemas!

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,182 @@ related:
...
+For all such properties, validation succeeds if the child instance validates against the `unevaluatedProperties` schema. This means that `properties`, `patternProperties`, `additionalProperties`, and all in-place applicators must be evaluated before this keyword can be evaluated.
+
+- The value of `unevaluatedProperties` must be a valid JSON Schema.
+- It considers annotations from `properties`, `patternProperties`, and `additionalProperties`, both as adjacent keywords and in subschemas of adjacent keywords.
+- This annotation affects the behavior of `unevaluatedProperties` in parent schemas.
+- `unevaluatedProperties` cannot see outside the schema to which it belongs.
+- This keyword is similar to `additionalProperties` except that it can recognize properties declared in subschemas.
+- The annotation result of this keyword is the set of instance property names validated by this keyword's subschema.
+
+## Examples
```
jviotti, line 46: For keywords like these that provide annotations, we should document annotations too. Ideally, we would have something like a `<instance-annotations>` Hugo partial but can be just a code block for now. For `unevaluatedProperties`, it will produce an annotation of `true` if it was applied to any keyword

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,182 @@ related:
...
+
+Validation with `unevaluatedProperties` applies only to the child values of instance names that do not appear in the `properties`, `patternProperties`, `additionalProperties`, or `unevaluatedProperties` annotation results that apply to the instance location being validated.
+
+For all such properties, validation succeeds if the child instance validates against the `unevaluatedProperties` schema. This means that `properties`, `patternProperties`, `additionalProperties`, and all in-place applicators must be evaluated before this keyword can be evaluated.
+
+- The value of `unevaluatedProperties` must be a valid JSON Schema.
+- It considers annotations from `properties`, `patternProperties`, and `additionalProperties`, both as adjacent keywords and in subschemas of adjacent keywords.
+- This annotation affects the behavior of `unevaluatedProperties` in parent schemas.
```
jviotti, line 41: This one is probably worth exemplifying

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,182 @@ related:
...
+- `unevaluatedProperties` cannot see outside the schema to which it belongs.
```
jviotti, line 42: Not sure what you mean here. Can you clarify?

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,182 @@ related:
...
+- This keyword is similar to `additionalProperties` except that it can recognize properties declared in subschemas.
```
jviotti, line 43: Well, not really. That's not the only difference. They actually have different semantics

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,182 @@ related:
...
+
+{{<instance-pass `An instance with no unevaluated properties is valid`>}}
+{ "foo": "foo", "bar": 36 }
+{{</instance-pass>}}
+
+{{<instance-fail `An instance with unevaluated properties is invalid`>}}
+{ "foo": "foo", "bar": 36, "fooBar": false }
+{{</instance-fail>}}
+
+* For the first instace, the annotation result of `properties` is [ "foo" ], and the annotation result of `patternProperties` is [ "bar" ]. No properties remain unevaluated; hence, the instance is considered valid.
```
jviotti, line 100: I think we need to represent annotation results better. Can you use code blocks that use standard output format objects instead?

content/2020-12/unevaluated/unevaluatedProperties.markdown
AgniveshChaubey: Already covered!

AgniveshChaubey: In example no  6.

content/2020-12/unevaluated/unevaluatedProperties.markdown
AgniveshChaubey: It means `unevaluatedProperties` does not considers annotation from its uncle schema.
For example:
```json
{
  "allOf": [
    {
      "additionalProperties": true
    },
    {
      "unevaluatedProperties": false
    }
  ]
}
```
Here, `unevaluatedProperties` of `/allOf/1` subschema cannot see the annotations produced by `/allOf/additionalProperties`

content/2020-12/unevaluated/unevaluatedProperties.markdown
AgniveshChaubey: Okay! removed it.

AgniveshChaubey:
Hi Juan,
I have addressed your comments in this PR. As per your suggestion, I have also added the annotation output along with the explanation. In the previous review, you mentioned that after completing `properties`, `patternProperties`, and `additionalProperties`, we could directly say that it applies to the keywords not covered by the annotations of those. However, since I have included all the annotations, do we still need to refer to those sections?

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,346 @@ related:
...
+If this keyword is applied to any instance element, it produces an annotation value of `true`.
+
+## Explanation
+
+First, let's understand what evaluation means. `unevaluatedProperties` considers annotations from `properties`, `patternProperties`, and `additionalProperties`, both as adjacent keywords and in subschemas of adjacent keywords. Additionally, `unevaluatedProperties` is affected by other `unevaluatedProperties` in nested schemas (if present). Each of these keywords will produce an annotation of the properties that they've evaluated.
+- For `properties`, it's just the listed keys intersected with what's in the instance.
+- For `patternProperties`, it's the list of properties that matches at least one of the regexes.
+- For `additionalProperties`, it's basically the rest of the properties.
+
+If any of these are in subschemas of adjacent keywords, and those subschemas fail validation, those annotations are dropped in that case. The effect is that those properties are not considered evaluated.
```
jviotti, line 38: The annotations are not dropped, they are just not considered for this specific case. I think your general explanation makes sense but I feel its a bit too verbose. Maybe a more concise way to explain this would be something along the following lines:

- `properties`, `patternProperties`, `additionalProperties`, and `unevaluatedProperties` emit annotations for the properties they validated against
- If `properties`, `patternProperties`, `additionalProperties`, or `unevaluatedProperties` emitted an annotation at the same instance location (independently of the schema location) for a certain property, that property is considered to have been evaluated
- By definition, the `unevaluatedProperties` keyword is always interpreted AFTER `properties`, `patternProperties`, `additionalProperties`
- As its name implies, `unevaluatedProperties` is an applicator that will apply to any object property not evaluated
 
Now that I write the above, I wonder if it be better to first have a section purely on what "evaluation" means. Once that's out of the picture, the explanation section here can just say the last item: `unevaluatedProperties` applies to properties not previously evaluated?

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,346 @@ related:
 Annotations
 -----------
 
-If this keyword is applied to any instance element, it produces an annotation
-value of `true`.
+If this keyword is applied to any instance element, it produces an annotation value of `true`.
```
jviotti, line 29: This is wrong actually (I never realized). It is the set of evaluated properties, like for `properties`. Can you fix this while you are at it too? 🙏🏻

content/2020-12/unevaluated/unevaluatedProperties.markdown
jviotti: I see. This keyword is one of the complex ones most people don't really understand. To be honest, I think a large part of it is that even the spec mentions things like "can see outside the schema" or "adjacent keywords", etc which confuse people.

I feel like explaining it in terms of annotations is so much simpler. We can say that if for an object, any related keyword emitted a property as an annotation (independently of the schema structure), then that property is evaluated. And that's it. Anything not evaluated is tested against `unevaluatedProperties`.

Then we avoid mentioning scoping, adjacency, and a lot of other confusing topics. What do you think?

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,346 @@ related:
...
+    "annotation": [ "fooBar" ]
+  },
+  // ...
+]
+{{</instance-annotation>}}
+
+{{<instance-fail `An instance with unevaluated properties that do not conform to the 'unevaluatedProperties' subschema is invalid`>}}
+{ "foo": "foo", "bar": 36, "fooBar": "string" }
+{{</instance-fail>}}
+For the above two instances, the annotation result of `properties` is [ "foo" ], and the annotation result of nested `patternProperties` is [ "bar" ]. The `unevaluatedProperties` recognizes the annotations from `properties` as well as `patternProperties` (as it can see through adjacent and nested applicators) and ensures that 'fooBar' remains unevaluated and its subschema applies to 'fooBar'.
```
jviotti, line 260:
```suggestion
For the above two instances, the annotation result of `properties` is [ "foo" ], and the annotation result of nested `patternProperties` is [ "bar" ]. The `unevaluatedProperties` recognizes the annotations from `properties` as well as `patternProperties` (as it can see through adjacent and nested applicators as only the produced annotations matter, not the schema structure) and ensures that 'fooBar' remains unevaluated and its subschema applies to 'fooBar'.
```

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,346 @@ related:
...
+  // ...
+  {
+    "valid": true,
+    "keywordLocation": "/properties",
+    "instanceLocation": "",
+    "annotation": [ "foo" ]
+  },
+  {
+    "valid": true,
+    "keywordLocation": "/allOf/patternProperties",
```
jviotti, line 243:
```suggestion
    "keywordLocation": "/allOf/0/patternProperties",
```

You are missing the index

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,346 @@ related:
...
+    "bar": {
+      "properties": {
+        "bar": { "type": "string" }
+      }
+    }
+  }
+}
+{{</schema>}}
+
+{{<instance-pass `An instance with no unevaluated proeprties is valid`>}}
```
jviotti, line 346:
```suggestion
{{<instance-pass `An instance with no unevaluated properties is valid`>}}
```

jviotti:
I left a few comments, but much better! Overall:

- I think we can simplify the explanation a bit. This is a complex keyword to explain, but I think describing it in terms of annotations (given that we show annotation results in the example) will make a lot of sense and be quite straightforward
- The examples are great
- One thing you are discussing here but not exemplifying is nested `unevaluatedProperties`. Maybe add a tiny example on this?

content/2020-12/unevaluated/unevaluatedProperties.markdown
AgniveshChaubey: Yes, that seems like a good idea. Let's then separate them into two sections

content/2020-12/unevaluated/unevaluatedProperties.markdown
AgniveshChaubey: Yes, explaining it in terms of annotations seems a lot simpler. Let me write it that way.

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,346 @@ related:
...
+    "valid": true,
+    "keywordLocation": "/$ref/properties",
+    "instanceLocation": "",
+    "annotation": [ "bar" ]
+  },
+  // ...
+]
+{{</instance-annotation>}}
+
+{{<instance-fail `An instance with unevaluated proeprties is invalid`>}}
```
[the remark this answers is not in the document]
AgniveshChaubey: 🤦‍♂️!
This used to happen to me a lot when I was documenting properties, patternProperties, and additionalProperties. Somehow, it became muscle memory, and my hand automatically typed 'proeprties' instead of 'properties.' 😅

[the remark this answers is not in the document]
jviotti: haha yeah, same happens to me. At some point I'll add a spellchecker linter to this project that would catch these things

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,381 @@ related:
...
 
-If this keyword is applied to any instance element, it produces an annotation
-value of `true`.
+The annotation result of this keyword is the set of instance property names validated by this keyword's subschema.
+
+## Evaluation
+
+First, let's understand what evaluation means. `unevaluatedProperties` considers annotations from `properties`, `patternProperties`, and `additionalProperties`, both as adjacent keywords and in subschemas of adjacent keywords. Additionally, `unevaluatedProperties` is affected by other `unevaluatedProperties` in nested schemas (if present).
+
+- `properties`, `patternProperties`, `additionalProperties`, and `unevaluatedProperties` emit annotations for the properties they validated against.
```
jviotti, line 35: This was more my braindump about it, but I think you should covert it into nicer prose rather than just copy-pasting it.

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,381 @@ related:
...
+- If `properties`, `patternProperties`, `additionalProperties`, or `unevaluatedProperties` emitted an annotation at the same instance location (independently of the schema location) for a certain property, that property is considered to have been evaluated.
+- By definition, the `unevaluatedProperties` keyword is always interpreted AFTER `properties`, `patternProperties`, `additionalProperties`.
+- As its name implies, `unevaluatedProperties` is an applicator that will apply to any object property not evaluated.
+
+## Explanation
+
+Validation with `unevaluatedProperties` applies only to the child values of instance names that do not appear in the `properties`, `patternProperties`, `additionalProperties`, or `unevaluatedProperties` annotation results that apply to the instance location being validated. For all such properties, validation succeeds if the child instance validates against the `unevaluatedProperties` schema.
```
jviotti, line 42: I like this!

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,381 @@ related:
...
+    "valid": true,
+    "keywordLocation": "/properties",
+    "instanceLocation": "",
+    "annotation": [ "foo" ]
+  },
+  {
+    "valid": true,
+    "keywordLocation": "/allOf/0/unevaluatedProperties",
+    "instanceLocation": "",
+    "annotation": [ "foo", "bar" ]
```
jviotti, line 402: Here `foo` wouldn't happen, as its covered by `/properties`

content/2020-12/unevaluated/unevaluatedProperties.markdown
```diff
@@ -26,5 +26,381 @@ related:
...
+    "keywordLocation": "/allOf/0/additionalProperties",
+    "instanceLocation": "",
+    "annotation": [ "foo", "bar" ]
+  },
+  // ...
+]
+{{</instance-annotation>}}
+
+* In the first case, there are no unevaluated properties.
+* In the second case, the nested `{ additionalProperties: true }` evaluated all the unevaluated properties. So there's nothing left unevaluated.
```
jviotti, line 323:
```suggestion
* In the second case, the nested `{ additionalProperties: true }` evaluated all the remaining properties. So there's nothing left unevaluated.
```

content/2020-12/unevaluated/unevaluatedProperties.markdown
AgniveshChaubey: Initially, I too had the same doubt. But when I checked with json-everything, its output at `/allOf/0/unevaluatedProperties` included both properties (i.e., foo and bar). Additionally, someday I asked a question in the Slack chat regarding the same, i.e, on what properties do `allOf`, `anyOf`, etc. apply?' (e.g., do they apply only to only unevaluated properties or to all properties)? And if I remember correctly, Greg replied that they should cover all the properties of an instance.

Also, if a certain property is covered by the `properties` subschema and, let's say, the same property also matches one of the `patternProperties` regexes, then the `patternProperties` annotation list also contains that property. The same should be true for `allOf`, `anyOf`, etc., as well. Isn't it?

jviotti: This is the relevant spec section for `unevaluatedProperties`: https://json-schema.org/draft/2020-12/json-schema-core#section-11.3, which says:

> The annotation result of this keyword is the set of instance property names validated by this keyword's subschema. 

So at least according to that description, if a property is not validated by `unevaluatedProperties`, then it shouldn't be included in the annotation results?

Note that `properties` and `patternProperties` can have an overlap on their annotation results, as they are both independent of each other, where as i.e. `additionalProperties` depends on the other two (hence no overlap).

Do you have a link to the JSON Schema Slack discussion you refer to and potentially a screenshot from JSON Everything? Maybe I'm reading the spec wrong, or there might be a little bug in JSON Everything that you can report!

AgniveshChaubey: Here is the link to that specific thread - https://json-schema.slack.com/archives/C5CF75URH/p1710097896347719. More specifically, checkout [this reply](https://json-schema.slack.com/archives/C5CF75URH/p1710102838738819?thread_ts=1710097896.347719&cid=C5CF75URH).

And here is the output from JSON Everything for the above example.
```json

{
  "valid": true,
  "evaluationPath": "",
  "schemaLocation": "https://json-everything.net/257432a882#",
  "instanceLocation": "",
...
  "annotations": {
    "properties": [
      "foo"
    ]
  },
...
  "details": [
    {
      "valid": true,
      "evaluationPath": "/properties/foo",
      "schemaLocation": "https://json-everything.net/257432a882#/properties/foo",
      "instanceLocation": "/foo"
    },
    {
      "valid": true,
      "evaluationPath": "/allOf/0",
      "schemaLocation": "https://json-everything.net/257432a882#/allOf/0",
      "instanceLocation": "",
...
      "annotations": {
        "unevaluatedProperties": [
          "foo",
          "bar"
        ]
      },
...
      "details": [
        {
          "valid": true,
          "evaluationPath": "/allOf/0/unevaluatedProperties",
          "schemaLocation": "https://json-everything.net/257432a882#/allOf/0/unevaluatedProperties",
          "instanceLocation": "/foo/foo"
        },
        {
          "valid": true,
          "evaluationPath": "/allOf/0/unevaluatedProperties",
          "schemaLocation": "https://json-everything.net/257432a882#/allOf/0/unevaluatedProperties",
          "instanceLocation": "/bar/bar"
        }
      ]
    }
  ]
}
```

AgniveshChaubey: In short, Greg explained that -

Schemas cannot see outside of themselves.  There scope is limited to their contents.  Thus `{ unevaluatedProperties: true }` evaluates all of the properties.

jviotti: Ah, makes sense. I think its fine then. No need for additional explanation.

jviotti (approved):

Issue #97, opened 2024-02-05
Document `maxProperties`


jviotti:

jviotti:
I guess this one is not that interesting, but we can show `maxProperties` alone (without `properties`, etc) and in conjunction with `properties`, `additionalProperties`, and `patternProperties`? Just to show that the latter don't affect this keyword

AgniveshChaubey:
This task was addressed in #129

PR #129, opened 2024-03-14
Keyword `maxProperties`

AgniveshChaubey:
This PR addresses #97

### DIFF
--- a/content/2020-12/validation/maxProperties.markdown
+++ b/content/2020-12/validation/maxProperties.markdown
@@ -26,7 +26,7 @@ The `maxProperties` keyword is used to specify the maximum number of properties 
 
 ## Examples
 
-{{<schema `Schema with the 'maxProperties' keyword`>}}
+{{<schema `Schema with 'maxProperties' keyword`>}}
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "maxProperties": 2
@@ -41,7 +41,7 @@ The `maxProperties` keyword is used to specify the maximum number of properties 
 { "foo": 3, "bar": "hi", "baz": true }
 {{</instance-fail>}}
 
-{{<schema `Schema with the 'maxProperties' keyword`>}}
+{{<schema `Schema with 'maxProperties' and 'properties' keywords`>}}
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "type": "object",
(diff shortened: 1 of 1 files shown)


content/2020-12/validation/maxProperties.markdown
```diff
@@ -18,3 +18,97 @@ related:
...
+{{</schema>}}
+
+{{<instance-fail `An instance without 'address' property is invalid`>}}
+{ "name": "John", "age": 42 }
+{{</instance-fail>}}
+
+{{<instance-fail `An instance with more than 2 properties is invalid`>}}
+{ "name": "John", "age": 42, "address": "some address" }
+{{</instance-fail>}}
+* _It is important to note that one should be super cautious when using the `required` and `maxProperties` keywords together in a schema because it can create a situation where the instance will always fail the validation, as shown in the above example._
```
jviotti, line 114:
```suggestion
* _It is important to note that one should be cautious when using the `required` and `maxProperties` keywords together in a schema because it can create a situation where the instance will always fail the validation, as shown in the above example._
```

Super feels a bit too informal?

content/2020-12/validation/maxProperties.markdown
```diff
@@ -18,3 +18,97 @@ related:
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "maxProperties": 2
+}
+{{</schema>}}
+
+{{<instance-pass `An instance with 2 or less properties is valid`>}}
+{ "foo": 3, "bar": "hi" }
+{{</instance-pass>}}
+
+{{<instance-fail `An instance with more than 2 properties is invalid`>}}
```
jviotti, line 40: As in `minProperties`, maybe add an example of a non-object instance?

content/2020-12/validation/maxProperties.markdown
```diff
@@ -18,3 +18,97 @@ related:
...
+{{<schema `Schema with 'maxProperties' and 'required' keywords`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "properties": {
+    "name": { "type": "string" },
+    "age": { "type": "integer" },
+    "address": { "type": "string" }
+  },
+  "required": [ "name", "age", "address" ],
```
jviotti, line 102: Haha good example!

content/2020-12/validation/maxProperties.markdown
jviotti: Great point. Another potential case on this would be combining `minProperties` and `maxProperties` where `minProperties` is higher than the other one

jviotti (approved):
Some comments, but looks good!

Issue #98, opened 2024-02-05
Document `minProperties`


jviotti:

jviotti:
Same as https://github.com/Intelligence-AI/learnjsonschema.com/issues/97

AgniveshChaubey:
This task was addressed in #130

PR #130, opened 2024-03-14
Keyword `minProperties`

AgniveshChaubey:
This PR addresses #98

### DIFF
--- a/content/2020-12/validation/minProperties.markdown
+++ b/content/2020-12/validation/minProperties.markdown
@@ -18,3 +18,66 @@ related:
   - vocabulary: applicator
     keyword: additionalProperties
 ---
+
+The `minProperties` keyword is used to specify the minimum number of properties allowed in an object instnace. If the number of properties in the object is less than the value specified by `minProperties`, the validation fails.
+* It applies specifically to object instances.
+* The value of this keyword must be a non-negative integer (0 or greater).
+* Omitting this keyword has the same behavior as a value of 0.
+
+## Examples
+
+{{<schema `Schema with 'minProperties' keyword`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "minProperties": 1
+}
+{{</schema>}}
+
+{{<instance-pass `An instance with 1 or more properties is valid`>}}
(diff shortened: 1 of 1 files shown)


content/2020-12/validation/minProperties.markdown
```diff
@@ -18,3 +18,66 @@ related:
...
+## Examples
+
+{{<schema `Schema with 'minProperties' keyword`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "minProperties": 1
+}
+{{</schema>}}
+
+{{<instance-pass `An instance with 1 or more properties is valid`>}}
```
jviotti, line 36: Maybe also show off a non-object instance, where the keyword is irrelevant?

content/2020-12/validation/minProperties.markdown
```diff
@@ -18,3 +18,66 @@ related:
...
+  "type": "object",
+  "patternProperties": {
+    "^[Aa]ge$": { "type": "integer" }
+  },
+  "additionalProperties": { "type": "string" },
+  "minProperties": 2
+}
+{{</schema>}}
+
+{{<instance-pass `An instance with 2 or more properties is valid`>}}
```
jviotti, line 77: I think another interesting instance example here would be two or more properties when non of them match `patternProperties`

jviotti (approved):
Left some minor suggestions but looks good

content/2020-12/validation/minProperties.markdown
```diff
@@ -18,3 +18,66 @@ related:
   - vocabulary: applicator
     keyword: additionalProperties
 ---
+
+The `minProperties` keyword is used to specify the minimum number of properties allowed in an object instnace. If the number of properties in the object is less than the value specified by `minProperties`, the validation fails.
```
jviotti, line 22:
```suggestion
The `minProperties` keyword is used to specify the inclusive minimum number of properties allowed in an object instnace. If the number of properties in the object is less than the value specified by `minProperties`, the validation fails.
```

Maybe just to clarify? Same on `maxProperties`

PR #135, opened 2024-03-31
Keyword `title`

AgniveshChaubey:
This PR addresses #71.

Please have a look at the last example and let me know whether I understood correctly what you meant to convey in issue #71.

Regarding the output format for annotations, I've maintained it as per the discussion in issue #133. Although you previously suggested replacing the ellipsis used in the annotation with the complete standard output, I haven't replaced it as the complete output would be too large. Please let me know your thoughts on this.

### DIFF
--- a/content/2020-12/meta-data/title.markdown
+++ b/content/2020-12/meta-data/title.markdown
@@ -51,7 +51,7 @@ The `title` keyword in JSON Schema is used to provide a human-readable label for
   "valid": true,
   "keywordLocation": "/title",
   "instanceLocation": "",
-  "annotations": "Age of a person"
+  "annotation": "Age of a person"
 }
 {{</instance-annotation>}}
 
@@ -99,49 +99,31 @@ The `title` keyword in JSON Schema is used to provide a human-readable label for
     "valid": true,
     "keywordLocation": "/title",
     "instanceLocation": "",
-    "annotations": "Personal Info"
-  },
-  {
-    "valid": true,
-    "keywordLocation": "/properties",
-    "instanceLocation": "",
-    "annotations": [ "name", "age" ]
+    "annotation": "Personal Info"
   },
   {
     "valid": true,
     "keywordLocation": "/if/title",
     "instanceLocation": "",
-    "annotations": "if block"
-  },
-  {
-    "valid": true,
-    "keywordLocation": "/if/properties",
-    "instanceLocation": "",
-    "annotations": [ "age" ]
+    "annotation": "if block"
   },
   {
     "valid": true,
     "keywordLocation": "/if/properties/age/title",
     "instanceLocation": "/age",
-    "annotations": "'if' true"
+    "annotation": "'if' true"
   },
   {
     "valid": true,
     "keywordLocation": "/then/title",
     "instanceLocation": "",
-    "annotations": "then block",
-  },
-  {
-    "valid": true,
-    "keywordLocation": "/then/properties",
-    "instanceLocation": "",
-    "annotations": [ "eligible" ]
+    "annotation": "then block",
   },
   {
     "valid": true,
     "keywordLocation": "/then/properties/eligible/title",
     "instanceLocation": "/eligible",
-    "annotations": "then applied"
+    "annotation": "then applied"
   },
   // ...
 ]
@@ -172,13 +154,13 @@ The `title` keyword in JSON Schema is used to provide a human-readable label for
     "valid": true,
     "keywordLocation": "/title",
     "instanceLocation": "",
-    "annotations": "Person's name"
+    "annotation": "Person's name"
   },
   {
     "valid": true,
     "keywordLocation": "/$ref/title",
     "instanceLocation": "",
-    "annotations": "Person's name"
+    "annotation": "Person's name"
   },
   // ...
 ]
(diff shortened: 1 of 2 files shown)


content/2020-12/meta-data/title.markdown
```diff
@@ -27,3 +27,165 @@ Annotations
...
+{{</schema>}}
+
+{{<instance-pass `An instance with a numeric value is valid`>}}
+45
+{{</instance-pass>}}
+
+{{<instance-annotation>}}
+{
+  "valid": true,
+  "keywordLocation": "/",
```
jviotti, line 52:
```suggestion
  "keywordLocation": "/title",
```

The keyword emitting the annotation is `title`

content/2020-12/meta-data/title.markdown
```diff
@@ -27,3 +27,165 @@ Annotations
...
+  "instanceLocation": "",
+  "annotations": {
```
jviotti, line 54:
```suggestion
  "annotations": "Age of a person"
```

And the annotation is the title itself

content/2020-12/meta-data/title.markdown
```diff
@@ -27,3 +27,165 @@ Annotations
...
+{{</instance-pass>}}
+
+{{<instance-annotation>}}
+[
+  /// ...
+  {
+    "valid": true,
+    "keywordLocation": "/",
+    "instanceLocation": "",
+    "annotations": {
```
jviotti, line 104: This annotation example is mixing up annotations from different keywords. Each keyword generates a separate annotation. So you would have one string annotation by the `/title` keyword, one annotation by the `properties` keyword, etc

content/2020-12/meta-data/title.markdown
```diff
@@ -27,3 +27,165 @@ Annotations
...
+  "eligible": true
+}
+{{</instance-pass>}}
+
+{{<instance-annotation>}}
+[
+  /// ...
+  {
+    "valid": true,
+    "keywordLocation": "/",
```
jviotti, line 102: Also, this pointer is here is invalid. `/` means "the empty property". For the top-level object, the correct pointer is just `""`

In any case, this whole thing is invalid, as the root of the schema itself is not producing the annotations. Each of the corresponding keywords is, with the corresponding `keywordLocation`

content/2020-12/meta-data/title.markdown
```diff
@@ -27,3 +27,165 @@ Annotations
...
+    "instanceLocation": "",
+    "annotations": {
+      "title": "Personal Info",
+      "properties": [ "name", "age" ],
+      "if": true
```
jviotti, line 107: I don't think `if` produces an annotation?

jviotti:
The last example is good, but the annotation outputs require correction. In many of these, you have a single output unit from a non-keyword (like `""`) producing multiple annotations for different keywords. We want each individual keyword to produce an annotation of its own as a separate output unit

content/2020-12/meta-data/title.markdown
AgniveshChaubey: I too was confused with it. This annotation result was produced by json-everything and I intentionally kept it to get your feedback on this.

AgniveshChaubey:
Updated the annotation results!

content/2020-12/meta-data/title.markdown
jviotti: Yeah, doesn't seem to be standard according to https://json-schema.org/draft/2020-12/json-schema-core#section-10.2.2.1. Let's remove it!

content/2020-12/meta-data/title.markdown
```diff
@@ -27,3 +27,159 @@ Annotations
...
+{{<instance-pass `An instance with a numeric value is valid`>}}
+45
+{{</instance-pass>}}
+
+{{<instance-annotation>}}
+{
+  "valid": true,
+  "keywordLocation": "/title",
+  "instanceLocation": "",
+  "annotations": "Age of a person"
```
jviotti, line 54: This should be just `annotation`

content/2020-12/meta-data/title.markdown
```diff
@@ -27,3 +27,159 @@ Annotations
...
+{{</instance-pass>}}
+
+{{<instance-annotation>}}
+[
+  /// ...
+  {
+    "valid": true,
+    "keywordLocation": "/title",
+    "instanceLocation": "",
+    "annotations": "Personal Info"
```
jviotti, line 102: Same in all of these. Should be `annotation` (in singular form)

content/2020-12/meta-data/title.markdown
```diff
@@ -27,3 +27,159 @@ Annotations
...
+  },
+  {
+    "valid": true,
+    "keywordLocation": "/properties",
```
jviotti, line 106: While this is correct, maybe let's use ellipses to "hide" any annotation that is not relevant to the keywords we are describing? i.e. here we can only show annotations generated by `title`?

AgniveshChaubey: Sure!

AgniveshChaubey:
Updated!

jviotti (approved):
Awesome!

Issue #48, opened 2024-02-05
Document `$schema`


jviotti:

jviotti:
- Explain that this refers to the "meta-schema" of a schema. Maybe explain what that means in a line?
- Explain that declaring `$schema` is VERY recommended. Definitely encourage users to do so, otherwise JSON Schema implementations might have to guess what version of JSON Schema the schema has been written for
- Explain that declaring `$schema` is also VERY important to know which vocabularies are set for the given schema. Otherwise, if you are using custom vocabularies, it could get ambiguous?
- Explain that the schema is expected to successfully validate against its own meta schema
- Also touch on JSON Schema bundling, where you may have nested `$schema` on the same schema resource?

PR #136, opened 2024-03-31
Keyword `$schema`

AgniveshChaubey:
This PR addresses #48.

### DIFF
--- a/content/2020-12/core/schema.markdown
+++ b/content/2020-12/core/schema.markdown
@@ -16,3 +16,52 @@ related:
   - vocabulary: core
     keyword: $defs
 ---
+
+The `$schema` keyword is a fundamental element in JSON Schema. It serves the two crucial purposes:
+1. **Dialect Identification:** It specifies the specific version (dialect) of JSON Schema the schema adheres to. This ensures implementations (tools and libraries) interpret the schema correctly based on the intended dialect's rules.
+
+2. **Meta-Schema Validation:** The value of `$schema` is a URI pointing to a "meta-schema", which defines the structure and validation rules for JSON Schemas. A schema that describes another schema is called a "meta-schema". The schema is expected to be valid against its own meta-schema.
+
+* The value of this keyword must be a [URI](https://json-schema.org/draft/2020-12/json-schema-core#RFC3986).
+* The current schema must be valid against the meta-schema identified by this URI.
+* The `$schema` keyword should be used in the document root schema object, and may be used in the root schema objects of embedded schema resources.
+* If this keyword is absent from the document root schema, the resulting behavior is implementation-defined.
+
+{{<alert>}}
+**Important:**
+* Declaring `$schema` is highly recommended for several reasons. It ensures clarity by explicitly stating the version of JSON Schema the schema follows. This helps JSON Schema implementations (tools and libraries) understand how to interpret and validate the schema accurately.
+* JSON Schema versions may introduce new keywords or modify existing ones. By specifying the `$schema`, you establish the specific vocabulary  that applies to your schema, preventing ambiguity, especially if you're using custom keywords.
+* The schema is expected to successfully validate against its own meta-schema, ensuring its correctness and adherence to the JSON Schema standard.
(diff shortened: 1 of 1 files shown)


content/2020-12/core/schema.markdown
```diff
@@ -16,3 +16,73 @@ related:
...
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "string"
+}
+{{</schema>}}
+
+{{<schema `Schema adhering to its meta-schema is valid`>}}
+{
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "items": [{ "type": "number" }]
```
jviotti, line 50:
```suggestion
  "items": [ { "type": "number" } ]
```

We use spaces within square brackets in other examples. Let's keep the spaces for consistency? Same applies to the examples below

content/2020-12/core/schema.markdown
```diff
@@ -16,3 +16,73 @@ related:
...
+}
+{{</schema>}}
+
+{{<schema `Schema not adhering to its meta-schema is invalid`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "items": [{ "type": "number" }]
+}
+{{</schema>}}
+* _The value of the `items` keyword can be either a valid JSON Schema or an array of valid JSON Schemas according to the 2019-09 dialect of JSON Schema. However, in the 2020-12 dialect of JSON Schema, the `items` keyword can only have a single valid JSON Schema. Therefore, setting the `items` keyword to an array of JSON Schemas makes it invalid according to the 2020-12 specification._
```
jviotti, line 60: We shouldn't be mentioning 2019-09. This documentation page is purely for 2020-12

content/2020-12/core/schema.markdown
AgniveshChaubey: Okay!

AgniveshChaubey:
Updated!

content/2020-12/core/schema.markdown
```diff
@@ -16,3 +16,73 @@ related:
...
+{{<schema `This declaration indicates that the schema is described by the JSON Schema 2020-12 dialect`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "string"
+}
+{{</schema>}}
+
+{{<schema `Schema adhering to its meta-schema is valid`>}}
+{
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
```
jviotti, line 49: I think we should omit 2019-09 examples, as this page is 2020-12 only. I think this whole schema can be removed? The one before and after are good examples already

content/2020-12/core/schema.markdown
```diff
@@ -16,3 +16,73 @@ related:
...
+}
+{{</schema>}}
+
+{{<schema `Schema not adhering to its meta-schema is invalid`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "items": [ { "type": "number" } ]
+}
+{{</schema>}}
+* _The value of the `items` keyword can be either a valid JSON Schema or an array of valid JSON Schemas according to the 2019-09 dialect of JSON Schema. However, in the 2020-12 dialect of JSON Schema, the `items` keyword can only have a single valid JSON Schema. Therefore, setting the `items` keyword to an array of JSON Schemas makes it invalid according to the 2020-12 specification._
```
jviotti, line 60: This feels like a very specific example that not many people might relate to. What about something more obvious like setting `properties` to an array? That way we also avoid mentioning 2019-09

content/2020-12/core/schema.markdown
AgniveshChaubey: I added it to demonstrate that when a newer spec changes the behavior of a particular keyword, it becomes crucial to specify the correct dialect if you want to retain the old behavior associated with that keyword. But yeah, it's a very specific case and may be less relatable to the general audience. Let's remove it."

content/2020-12/core/schema.markdown
AgniveshChaubey: Done!

jviotti (approved):

Issue #51, opened 2024-02-05
Document `$comment`


jviotti:

jviotti:
I guess just explain that this does nothing. No annotations and no validation results. Just as a standard place to put arbitrary text for readability

AgniveshChaubey:
This task was addressed in #137,

PR #137, opened 2024-03-31
Keyword `$comment`

AgniveshChaubey:
This PR addresses #51

### DIFF
--- a/content/2020-12/core/comment.markdown
+++ b/content/2020-12/core/comment.markdown
@@ -10,10 +10,11 @@ introduced_in: draft7
 index: -9
 ---
 
-The `$comment` keyword is used to provide a way for schema authors to include explanatory or clarifying comments within the schema itself. However, it's important to note that JSON Schema implementations are not required to attach any specific meaning or behavior to `$comment`, and they may even ignore or strip it out entirely. Therefore, `$comment` is primarily useful for leaving notes to future editors of the schema rather than communicating with users of the schema.
+The `$comment` keyword is used to provide a way for schema authors to include explanatory or clarifying comments within the schema itself. However, it's important to note that JSON Schema implementations are required to NOT add meaning to `$comment`, and not even collect it as an annotation. Additionally, they may even ignore or strip it out entirely. Therefore, `$comment` is primarily useful for leaving notes to future editors of the schema rather than communicating with users of the schema.
 
 * The value of this keyword must be a string.
 * Implementations ignore `$comment` during validation.
+* It will not even be collected as an annotation.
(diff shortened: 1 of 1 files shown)


content/2020-12/core/comment.markdown
```diff
@@ -9,3 +9,44 @@ metaschema: "https://json-schema.org/draft/2020-12/meta/core"
 introduced_in: draft7
 index: -9
 ---
+
+The `$comment` keyword is used to provide a way for schema authors to include explanatory or clarifying comments within the schema itself. However, it's important to note that JSON Schema implementations are not required to attach any specific meaning or behavior to `$comment`, and they may even ignore or strip it out entirely. Therefore, `$comment` is primarily useful for leaving notes to future editors of the schema rather than communicating with users of the schema.
```
jviotti, line 13: > However, it's important to note that JSON Schema implementations are not required to attach any specific meaning or behavior to `$comment`, and they may even ignore or strip it out entirely.

Its not that they are not required to attach meaning, but that they are required to NOT add meaning to it, and not even collect it as an annotation

content/2020-12/core/comment.markdown
```diff
@@ -9,3 +9,44 @@ metaschema: "https://json-schema.org/draft/2020-12/meta/core"
...
+
+* The value of this keyword must be a string.
+* Implementations ignore `$comment` during validation.
```
jviotti, line 16: Also clarify it will not even be collected as an annotation.

content/2020-12/core/comment.markdown
```diff
@@ -9,3 +9,44 @@ metaschema: "https://json-schema.org/draft/2020-12/meta/core"
...
+* It can be placed anywhere within a JSON Schema to provide additional context or explanation.
+* Some tools might even remove this keywords for size optimization.
```
jviotti, line 18: I never saw this before. Do you have any links?

AgniveshChaubey: I found this point mentioned here [here](https://json-schema.org/draft/2020-12/json-schema-core#section-8.3), so included it in this documentation as well!

> Implementations MAY strip "$comment" values at any point during processing. In particular, this allows for shortening schemas when the size of deployed schemas is a concern.

jviotti: Ah, interesting. I guess its good to keep it, but maybe make it a link to that part of the spec for future reference?

AgniveshChaubey: Okay

AgniveshChaubey:
Done!

Issue #78, opened 2024-02-05
Describe `format` (annotation)


jviotti:

jviotti:
Oh, and we should create a table summarizing the possible values of `format` like we do for the `type` keyword

AgniveshChaubey:
This task was addressed in #138.

PR #138, opened 2024-03-31
Keyword `format` (annotation)

AgniveshChaubey:
This PR addresses #78.

### DIFF
--- a/content/2020-12/format-annotation/format.markdown
+++ b/content/2020-12/format-annotation/format.markdown
@@ -75,6 +75,19 @@ When using `format` from Format Annotation, it's recommended that you provide yo
 45
 {{</instance-pass>}}
 
+{{<instance-annotation>}}
+[
+  // ...
+  {
+    "valid": true,
+    "keywordLocation": "/format",
+    "instanceLocation": "",
+    "annotation": "email"
+  },
+  // ...
+]
+{{</instance-annotation>}}
+
 {{<schema `Schema with the 'format' keyword having validation rules for email`>}}
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
@@ -94,3 +107,16 @@ When using `format` from Format Annotation, it's recommended that you provide yo
 {{<instance-pass `'format' keyword is irrelevant for instances with values other than strings`>}}
 true
 {{</instance-pass>}}
+
+{{<instance-annotation>}}
+[
+  // ...
+  {
+    "valid": true,
+    "keywordLocation": "/format",
+    "instanceLocation": "",
+    "annotation": "email"
+  },
+  // ...
+]
+{{</instance-annotation>}}
--- a/content/2020-12/format-annotation/format.markdown
+++ b/content/2020-12/format-annotation/format.markdown
@@ -52,8 +52,7 @@ When using `format` from Format Annotation, it's recommended that you provide yo
 * `It allows for the semantic identification of certain kinds of string values. For instance, it can indicate that a string value should be interpreted as a date, email, URI, etc.
 * `format` is solely an annotation and does not enforce any validation. It's meant to provide information about the expected format of the string.
 * Implementations may choose to enable format as an assertion, meaning that validation fails if the value doesn't conform to the specified format semantics. However, this is not mandatory and must be explicitly enabled.
-* While users can define and use their own custom `formats` (e.g., "format": "foobar"), it's recommended to refrain from overloading the format keyword for future compatibility reasons. Instead, define custom keywords for specific validation requirements.
-* In the event that you define your own "foobar" and JSON Schema subsequently chooses to define "foobar," you may encounter difficulties.
+* While users can define and use their own custom `formats` (e.g., "format": "foobar"), it's recommended to refrain from overloading the format keyword for future compatibility reasons. Instead, define custom keywords for specific validation requirements. For example in the event that you define your own "foobar" and JSON Schema subsequently chooses to define "foobar," you may encounter difficulties.
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


content/2020-12/format-annotation/format.markdown
```diff
@@ -41,3 +41,56 @@ Defined Formats
 | `"json-pointer"`          | JSON Pointer         | https://json-schema.org/draft/2020-12/json-schema-validation.html#section-7.3.7 |
 | `"relative-json-pointer"` | JSON Pointer         | https://json-schema.org/draft/2020-12/json-schema-validation.html#section-7.3.7 |
 | `"regex"`                 | Regular Expressions  | https://json-schema.org/draft/2020-12/json-schema-validation.html#section-7.3.8 |
+
+## Explanation
+
+The `format` keyword in JSON Schema's Format Annotation vocabulary serves to provide basic semantic identification of certain types of string values. Compared to older JSON Schema drafts, the `format` keyword in the official 2020-12 dialect is purely an annotation meant to be informative and to carry semantics rather than to perform any validation.
```
jviotti, line 47:
```suggestion
The `format` keyword in JSON Schema's Format Annotation vocabulary serves to provide basic semantic identification of certain types of string values. Compared to older JSON Schema versions, the `format` keyword in the official 2020-12 dialect is purely an annotation meant to be informative and to carry semantics rather than to perform any validation.
```

content/2020-12/format-annotation/format.markdown
```diff
@@ -41,3 +41,56 @@ Defined Formats
...
+
+When using `format` from Format Annotation, it's expected that you provide your validation rules alongside the `format`. The implementation may choose to treat `format` as an assertion and attempt to validate the value's conformance to the specified semantics. However, this behavior must be explicitly enabled and is typically disabled by default. Implementations should document their level of support for such validation.
```
jviotti, line 49: > it's expected that you provide your validation rules alongside the `format`

Let's say its recommended.

> The implementation may choose to treat `format` as an assertion and attempt to validate the value's conformance to the specified semantics. However, this behavior must be explicitly enabled and is typically disabled by default. Implementations should document their level of support for such validation.

This is not correct. An implementation that allows you optionally treat them as validation is non compliant. If you are using Format Annotation, the implementation MUST NOT treat that as a validation constraint under any circumstance. If you want assertions, you should use Format Assertion

content/2020-12/format-annotation/format.markdown
```diff
@@ -41,3 +41,56 @@ Defined Formats
...
+
+* `format` applies only to string types, describing logical formats for these strings.
+* `It allows for the semantic identification of certain kinds of string values. For instance, it can indicate that a string value should be interpreted as a date, email, URI, etc.
+* `format` is solely an annotation and does not enforce any validation. It's meant to provide information about the expected format of the string.
+* Implementations may choose to enable format as an assertion, meaning that validation fails if the value doesn't conform to the specified format semantics. However, this is not mandatory and must be explicitly enabled.
```
jviotti, line 54: Again, this would be non-compliant

content/2020-12/format-annotation/format.markdown
```diff
@@ -41,3 +41,56 @@ Defined Formats
...
+* While users can define and use their own custom `formats` (e.g., "format": "foobar"), it's recommended to refrain from overloading the format keyword for future compatibility reasons. Instead, define custom keywords for specific validation requirements.
```
jviotti, line 55: This is a good point! You can even explain further saying that if you define your own `foobar` and JSON Schema later decides to define `foobar`, you will be in trouble

content/2020-12/format-annotation/format.markdown
AgniveshChaubey: > This is not correct. An implementation that allows you optionally treat them as validation is non compliant. If you are using Format Annotation, the implementation MUST NOT treat that as a validation constraint under any circumstance. If you want assertions, you should use Format Assertion

But, the [spec](https://json-schema.org/draft/2020-12/json-schema-validation#section-7.2.1) says implementations can optionally choose to do so.

> Implementations MAY still treat "format" as an assertion in addition to an annotation and attempt to validate the value's conformance to the specified semantics.

jviotti: Ah, interesting. I didn't realize that. Fine then, but can you link this part to the spec?

content/2020-12/format-annotation/format.markdown
```diff
@@ -41,3 +41,56 @@ Defined Formats
...
+
+* `format` applies only to string types, describing logical formats for these strings.
+* `It allows for the semantic identification of certain kinds of string values. For instance, it can indicate that a string value should be interpreted as a date, email, URI, etc.
+* `format` is solely an annotation and does not enforce any validation. It's meant to provide information about the expected format of the string.
+* Implementations may choose to enable format as an assertion, meaning that validation fails if the value doesn't conform to the specified format semantics. However, this is not mandatory and must be explicitly enabled.
+* While users can define and use their own custom `formats` (e.g., "format": "foobar"), it's recommended to refrain from overloading the format keyword for future compatibility reasons. Instead, define custom keywords for specific validation requirements. For example in the event that you define your own "foobar" and JSON Schema subsequently chooses to define "foobar," you may encounter difficulties.
+
+## Examples
```
jviotti, line 57: I just checked given the other PRs and `format` is collected as an annotation. Can you add annotation examples here too?

jviotti (changes requested):

AgniveshChaubey:
Annotation added!

jviotti (approved):

Issue #79, opened 2024-02-05
Describe `format` (assertion)


jviotti:

jviotti:
We should clarify well that this vocabulary is not used by default on the official 2020-12 dialect, and if you want to use it, you need to define your own custom dialect that pulls it in.

Also clarify that while this keyword in theory provides interoperable logical string type validation, a lot of implementations out there will not actually implement this vocabulary. Hence, suggest `format` from Format Annotation (which is available out of the box) alongside your custom validation in there.

Also clarify that `format` applies to only string types, as it is meant to describe "logical" formats for string types.

For examples:

- Define a custom meta-schema with `$vocabulary` that is the same as the official 2020-12 but with the Format Assertion vocabulary, and show i.e. `format: email` alongside some valid/invalid instances

jviotti:
Oh, and we should create a table summarizing the possible values of `format` like we do for the `type` keyword

PR #139, opened 2024-03-31
Keyword `format` (assertion)

AgniveshChaubey:
This PR addresses #79

### DIFF
--- a/content/2020-12/format-assertion/format.markdown
+++ b/content/2020-12/format-assertion/format.markdown
@@ -84,3 +84,16 @@ While the `format` keyword theoretically provides interoperable logical string t
 {{<instance-pass `'format' keyword is irrelevant for instances with values other than strings`>}}
 45
 {{</instance-pass>}}
+
+{{<instance-annotation>}}
+[
+  // ...
+  {
+    "valid": true,
+    "keywordLocation": "/format",
+    "instanceLocation": "",
+    "annotation": "email"
+  },
+  // ...
+]
+{{</instance-annotation>}}
--- a/content/2020-12/format-assertion/format.markdown
+++ b/content/2020-12/format-assertion/format.markdown
@@ -41,3 +41,34 @@ Defined Formats
 | `"json-pointer"`          | JSON Pointer         | https://json-schema.org/draft/2020-12/json-schema-validation.html#section-7.3.7 |
 | `"relative-json-pointer"` | JSON Pointer         | https://json-schema.org/draft/2020-12/json-schema-validation.html#section-7.3.7 |
 | `"regex"`                 | Regular Expressions  | https://json-schema.org/draft/2020-12/json-schema-validation.html#section-7.3.8 |
+
+## Explanation
+
+The `format` keyword of the "format-assertion" vocabulary allows for basic semantic identification of certain kinds of string values that are commonly used. It provides a way to specify logical formats for string types, such as dates, email addresses, URIs, etc. However, it's important to note that this vocabulary is not used by default in the official 2020-12 dialect of JSON Schema. If you want to utilize it, you would need to define your own custom dialect that includes this vocabulary.
+
+While the `format` keyword theoretically provides interoperable logical string type validation, many existing implementations may not support this vocabulary. Therefore, it's recommended to use the `format` keyword from the Format Annotation vocabulary (which is available out of the box) alongside any custom validation within the schema.
+
+## Examples
+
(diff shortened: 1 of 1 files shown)


content/2020-12/format-assertion/format.markdown
```diff
@@ -41,3 +41,34 @@ Defined Formats
...
+
+While the `format` keyword theoretically provides interoperable logical string type validation, many existing implementations may not support this vocabulary. Therefore, it's recommended to use the `format` keyword from the Format Annotation vocabulary (which is available out of the box) alongside any custom validation within the schema.
+
+## Examples
+
+{{<schema `Schema including the 'Format Assertion' vocabulary`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://json-schema.org/draft/2020-12/schema",
+  "$vocabulary": {
```
jviotti, line 57: This is not correct. `$vocabulary` will not apply to instances of the schema directly. You need to define a separate schema with `$vocabulary` and use that in `$schema` of another schema to make it work

AgniveshChaubey: I have negligible experience dealing with custom vocabularies. I tried this example with JSON-everything and it surprisingly it worked! That's why I added it here 😅.

You mean something like this?

meta-schema:
```json
{
  "$schema": "https://example.com/custom-meta-schema",
  "$id": "https://example.com/custom-meta-schema",
  "$vocabulary": {
    "https://json-schema.org/draft/2020-12/vocab/format-assertion": true
  },
  "allOf": [
    { "$ref": "https://json-schema.org/draft/2019-06/schema#" }
  ]
}
```

schema:
```json
{
  "$schema": "https://example.com/custom-meta-schema",
  "$id": "https://example.com/schema",
  "format": "email"
}
```

jviotti: Correct, but in every meta-schema you must always import the "core" vocabulary (the lack of it is considered an error). Plus instead of validating against the entirety of https://json-schema.org/draft/2020-12/schema in `allOf`, you can validate against the specific vocabulary meta-schemas: https://json-schema.org/draft/2020-12/meta/core and https://json-schema.org/draft/2020-12/meta/format-assertion.

Interesting that the other one work on JSON Everything. I believe the lack of consistency between implementations is a key problem in JSON Schema. Hard to know what's implementation-dependent and what's not for the average user

AgniveshChaubey:
Updated the example by creating a separate `meta-schema` and referencing it from another schema. Kindly check it and let me know if it is correct.

content/2020-12/format-assertion/format.markdown
AgniveshChaubey: Aren't there any standard set of rules that implementers must follow when writing their implementation to maintain consistency?

jviotti: They are, but they are often a bit loose IMO. Also, vocabularies are an experimental feature not fully fledged out (which carries its own problem!)

content/2020-12/format-assertion/format.markdown
```diff
@@ -41,3 +41,46 @@ Defined Formats
...
 | `"relative-json-pointer"` | JSON Pointer         | https://json-schema.org/draft/2020-12/json-schema-validation.html#section-7.3.7 |
 | `"regex"`                 | Regular Expressions  | https://json-schema.org/draft/2020-12/json-schema-validation.html#section-7.3.8 |
+
+## Explanation
+
+The `format` keyword of the "format-assertion" vocabulary allows for basic semantic identification of certain kinds of string values that are commonly used. It provides a way to specify logical formats for string types, such as dates, email addresses, URIs, etc. However, it's important to note that this vocabulary is not used by default in the official 2020-12 dialect of JSON Schema. If you want to utilize it, you would need to define your own custom dialect that includes this vocabulary.
+
+While the `format` keyword theoretically provides interoperable logical string type validation, many existing implementations may not support this vocabulary. Therefore, it's recommended to use the `format` keyword from the Format Annotation vocabulary (which is available out of the box) alongside any custom validation within the schema.
+
+## Examples
```
jviotti, line 51: Example looks good, but as I commented in the other `format` PR, `format` emits an annotation. Can we add that to the examples?

jviotti (changes requested):

AgniveshChaubey:
Annotation added!

jviotti (approved):

Issue #67, opened 2024-02-05
Document `propertyNames`


jviotti:

jviotti:
- Note that this schema only makes sense to apply to strings. Passing a schema here that matches something else would be invalid
- Maybe explain that logically this keyword defaults to `true`?
- Some valid/invalid examples here too

AgniveshChaubey:
This task was addressed in #142.

PR #142, opened 2024-04-02
Keyword `propertyNames`

AgniveshChaubey:
This PR addresses #67


content/2020-12/applicator/propertyNames.markdown
```diff
@@ -28,3 +28,62 @@ related:
   - vocabulary: unevaluated
     keyword: unevaluatedProperties
 ---
+
+The `propertyNames` keyword in is used to define constraints on the property names within an object instance. It allows you to specify a schema that all the property names in an object instance must adhere to.
+
+* The value of `propertyNames` must be a valid JSON Schema.
+* This keyword has no effect on instances other than objects.
```
jviotti, line 35: I'd even go further and clarify that according to the JSON grammar, object keys can only be strings, so if your schema matches i.e. objects, then it will always fail.

jviotti: Ah lol you literally said this in the alert below

Issue #57, opened 2024-02-05
Document `anyOf`


jviotti:

jviotti:
We can showcase 

- `anyOf` with one subschema only
- `anyOf ` with multiple subschemas
- `anyOf ` with boolean schemas?
- Nested `anyOf `?
- How schemas within `anyOf` interoperate with keywords that affect their siblings, like `additionalProperties` and `minContains`/`maxContains`

jviotti:
I guess we should also cover the case where there is more than one subschema that matches in `anyOf`

jviotti:
We can also talk about annotation collection might impact this. For example, if you are running with validation only, a JSON Schema implementation may stop validating `anyOf` when it finds a successful one, where as when collecting annotations it has to evaluate all of them. My book explains it like this:

<img width="669" alt="Screenshot 2024-02-26 at 19 46 43" src="https://github.com/Intelligence-AI/learnjsonschema.com/assets/2192773/eaf0cc65-d883-4f74-b038-48f411f7d1cf">

AgniveshChaubey:
This task was addressed in #153.

PR #153, opened 2024-04-13
Keyword `anyOf`

AgniveshChaubey:
This PR addresses #57.

### DIFF
--- a/content/2020-12/applicator/anyOf.markdown
+++ b/content/2020-12/applicator/anyOf.markdown
@@ -27,7 +27,7 @@ The `anyOf` keyword in JSON Schema is used to specify that an instance must vali
 
 * The value of this keyword must be a non-empty array.
 * Each item of the array must be a valid JSON Schema.
-anyOf
+
 ## Examples
 
 {{<schema `Schema with 'anyOf' keyword containing only one subschema`>}}
@@ -131,7 +131,7 @@ anyOf
 {{<instance-pass `An instance not conforming to the second subschema of 'anyOf' is also valid`>}}
 { "foo": true }
 {{</instance-pass>}}
-* _Remember, if any subschema within the `anyOf` keyword passes validation or has a boolean `true` value, the overall validation against `anyOf` stops at that point, indicating successful validation for this keyword._
+* _Remember, if any subschema within the `anyOf` keyword passes validation or has a boolean `true` value, the overall result of `anyOf` is considered valid._
 
 {{<schema `Schema with nested 'anyOf'`>}}
 {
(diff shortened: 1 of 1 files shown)


content/2020-12/applicator/anyOf.markdown
```diff
@@ -22,3 +22,149 @@ related:
   - vocabulary: applicator
     keyword: not
 ---
+
+The `anyOf` keyword in JSON Schema is used to specify that an instance must validate against at least one of the schemas provided in an array. It allows you to define multiple schemas, and if the data validates against any one of them, the validation passes.
+
+* The value of this keyword must be a non-empty array.
+* Each item of the array must be a valid JSON Schema.
+anyOf
```
jviotti, line 30: I think this line is a typo?

content/2020-12/applicator/anyOf.markdown
```diff
@@ -22,3 +22,149 @@ related:
...
+{{</schema>}}
+
+{{<instance-pass `An instance conforming to all the subschemas of 'anyOf' is valid`>}}
+{ "foo": "foo" }
+{{</instance-pass>}}
+
+{{<instance-pass `An instance not conforming to the second subschema of 'anyOf' is also valid`>}}
+{ "foo": true }
+{{</instance-pass>}}
+* _Remember, if any subschema within the `anyOf` keyword passes validation or has a boolean `true` value, the overall validation against `anyOf` stops at that point, indicating successful validation for this keyword._
```
jviotti, line 134: > the overall validation against `anyOf` stops at that point

This is not entirely correct. If collecting annotations, implementations will have to evaluate all of the branches, even after one matched. Even when not collecting annotations, the spec doesn't require implementations to stop after a first match. A badly designed implementation could still evaluate all of them

content/2020-12/applicator/anyOf.markdown
```diff
@@ -22,3 +22,149 @@ related:
...
+{{</schema>}}
+
+{{<instance-pass `An instance conforming to all the subschemas including the nested 'anyOf' is valid`>}}
+25
+{{</instance-pass>}}
+
+{{<instance-pass `An instance not conforming to the second subschema of top-level 'anyOf' is also valid`>}}
+10
+{{</instance-pass>}}
+* _For the first instance above, validation passes against the first subschema within `anyOf`, causing validation to halt at that point. Consequently, the `anyOf` is deemed valid without further validation against the second subschema._
```
jviotti, line 161: > causing validation to halt at that point

Same here. Let's not suggest validation halts. It may or may not!

content/2020-12/applicator/anyOf.markdown
jviotti: Ah, I see you explain this below. I guess here you can then say that as soon as one subschema succeeds, the result is valid (not indicating whether validation will stop or not)

content/2020-12/applicator/anyOf.markdown
AgniveshChaubey: Yup!

AgniveshChaubey:
Updated!

jviotti (approved):

PR #159, opened 2024-04-24
Keyword `default`

AgniveshChaubey:
This PR addresses #73.

### DIFF
--- a/content/2020-12/meta-data/default.markdown
+++ b/content/2020-12/meta-data/default.markdown
@@ -61,14 +61,14 @@ _**Note:** While it is recommended that the default value validate against its s
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "properties": {
     "name": { "type": "string", "default": "John" },
-    "education": {
+    "qualification": {
       "enum": [ "degree", "diploma" ],
       "default": "diploma"
     }
   },
   "if": {
     "properties": {
-      "education": { "const": "degree" }
+      "qualification": { "const": "degree" }
     }
   },
   "then": {
@@ -82,12 +82,12 @@ _**Note:** While it is recommended that the default value validate against its s
   },
   "else": {
     "properties": {
-      "diploma_certificate": {
+      "diplomaCertificate": {
         "type": "string",
         "default": "YYYYYY"
       }
     },
-    "required": [ "diploma_certificate" ]
+    "required": [ "diplomaCertificate" ]
   }
 }
 {{</schema>}}
@@ -95,7 +95,7 @@ _**Note:** While it is recommended that the default value validate against its s
 {{<instance-pass `An instance conforming to the schema is valid`>}}
 {
   "name": "Doe",
-  "qulaification": "degree",
+  "qualification": "degree",
   "degreeCertificate": "XXYYZZ"
 }
 {{</instance-pass>}}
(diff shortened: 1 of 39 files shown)


content/2020-12/meta-data/default.markdown
```diff
@@ -26,3 +26,138 @@ Annotations
 -----------
 
 This keyword produces the default value as the annotation value.
+
+## Explanation
+
+The `default` keyword in JSON Schema is used to specify a default value for an instance or its parts. This value is not used to fill in missing values during the validation process but can be used by non-validation tools such as documentation generators or form generators to provide hints to users.
```
jviotti, line 32:
```suggestion
The `default` keyword in JSON Schema is used to specify a default value for an instance. This value is not automatically used to fill in missing values during the validation process but can be used by tools such as documentation or form generators.
```

content/2020-12/meta-data/default.markdown
```diff
@@ -26,3 +26,138 @@ Annotations
...
+
+_**Note:** While it is recommended that the default value validate against its subschema, this requirement is not strictly enforced. However, it is strongly advised for people to ensure validation compatibility_
```
jviotti, line 34:
```suggestion
_**Note:** While it is recommended that the default value validate against its subschema, this requirement is not strictly enforced. 
```

I think the second sentence just says the same thing

content/2020-12/meta-data/default.markdown
```diff
@@ -26,3 +26,138 @@ Annotations
...
+    }
+  },
+  "if": {
+    "properties": {
+      "education": { "const": "degree" }
+    }
+  },
+  "then": {
+    "properties": {
+      "degree_certificate": {
```
jviotti, line 76: Can we use camelCase, as that's the style JSON Schema encourages?

content/2020-12/meta-data/default.markdown
```diff
@@ -26,3 +26,138 @@ Annotations
...
+  "if": {
+    "properties": {
+      "qualification": { "const": "degree" }
+    }
+  },
+  "then": {
+    "properties": {
+      "degreeCertificate": {
+        "type": "string",
+        "default": "XXXXXX"
```
jviotti, line 78: These placeholders feel a bit odd. Maybe generate a random short alphanumeric string instead? Something like `B0B8RKEZ90`? Would feel at least a bit more plausible

content/2020-12/meta-data/default.markdown
```diff
@@ -26,3 +26,138 @@ Annotations
...
+      }
+    },
+    "required": [ "degreeCertificate" ]
+  },
+  "else": {
+    "properties": {
+      "diplomaCertificate": {
+        "type": "string",
+        "default": "YYYYYY"
```
jviotti, line 87: Same here. Maybe `PW458C468E`?

content/2020-12/meta-data/default.markdown
```diff
@@ -26,3 +26,138 @@ Annotations
...
+    "required": [ "diplomaCertificate" ]
+  }
+}
+{{</schema>}}
+
+{{<instance-pass `An instance conforming to the schema is valid`>}}
+{
+  "name": "Doe",
+  "qualification": "degree",
+  "degreeCertificate": "XXYYZZ"
```
jviotti, line 99: `O5CYPZACTN`?

jviotti (approved):
I left a few comments regarding the placeholders, but looks good given that

content/2020-12/meta-data/default.markdown
AgniveshChaubey: Make sense!

Issue #36, opened 2023-10-23
Consider making the sidebar toggleable to *not* break down keywords by vocabulary


Julian:
I would love it if the sidebar was *not* broken down by vocabulary and simply listed all keywords in alphabetical order. That'd make it way easier to find a keyword when you know it exists. E.g. just now in ~<1 second I wanted to find both `dependentSchemas` and `dependentRequired`, and could not see why they weren't both there -- they were of course, but one was in one vocabulary and one in another, which took a few additional seconds to realize when simply scanning the page.

The counter to doing away entirely with vocabularies in the sidebar is that finding keywords by "relationship" to others -- which is kind of what vocabularies are trying to model -- then becomes harder.

Perhaps a meaningful way of solving both is to have the sidebar have 2 tabs -- one where keywords are broken down by vocabulary, and one where they're simply listed alphabetically. Or a radio slider which you could toggle from "by vocabulary" to "sorted" or whatever.

You may not be surprised on this opinion of mine :D -- feel free to close this of course if you think it's overboard!

jviotti:
> Perhaps a meaningful way of solving both is to have the sidebar have 2 tabs -- one where keywords are broken down by vocabulary, and one where they're simply listed alphabetically. Or a radio slider which you could toggle from "by vocabulary" to "sorted" or whatever.

I like this tab idea. Adding to my TODO list :)

jviotti:
Maybe some tab buttons at the top that say:

- By vocabulary (what we have now)
- By dialect (all at once as you want them)

jviotti:
Fixed by https://github.com/Intelligence-AI/learnjsonschema.com/pull/254

PR #162, opened 2024-04-26
First draft commit including 2019-09 files

Era-cell:
I have just copied and pasted 2020-12 folder as it is by renaming it.
Styling to be done on discussion
  😁

and, ignore the 2019-09 files, I have major changes in sidebar.html

### DIFF
--- a/content/2019-09/_index.md
+++ b/content/2019-09/_index.md
@@ -1,0 +1,24 @@
+---
+title: "2019-09"
+description: "JSON Schema: A Media Type for Describing JSON Documents"
+summary: "JSON Schema 2020-12 is a JSON media type for defining the structure of JSON data. JSON Schema is intended to define validation, documentation, hyperlink navigation, and interaction control of JSON data."
+organization: JSON Schema
+specification: "https://json-schema.org/draft/2020-12/json-schema-core.html"
+authors:
+  - github: awwright
+    first: Austin
+    last: Wright
+  - github: handrews
+    first: Henry
+    last: Andrews
+  - github: relequestual
+    first: Ben
+    last: Hutton
+  - github: gregsdennis
+    first: Greg
+    last: Dennis
+metaschema: "https://json-schema.org/draft/2020-12/schema"
+year: 2020
+bowtie: "https://bowtie.report/#/draft2020-12"
+---
+
--- a/content/2019-09/applicator/_index.markdown
+++ b/content/2019-09/applicator/_index.markdown
@@ -1,0 +1,25 @@
+---
+title: "Applicator"
+summary: "A vocabulary that defines applicator keywords that are recommended for use as the basis of other vocabularies."
+specification: "https://json-schema.org/draft/2020-12/json-schema-core.html#section-10"
+authors:
+  - github: awwright
+    first: Austin
+    last: Wright
+  - github: handrews
+    first: Henry
+    last: Andrews
+  - github: relequestual
+    first: Ben
+    last: Hutton
+  - github: gregsdennis
+    first: Greg
+    last: Dennis
+uri: "https://json-schema.org/draft/2020-12/vocab/applicator"
+metaschema: "https://json-schema.org/draft/2020-12/meta/applicator"
+official: true
+default: true
+index: -9999
+related:
+  - unevaluated
+---
--- a/content/2019-09/applicator/additionalProperties.markdown
+++ b/content/2019-09/applicator/additionalProperties.markdown
@@ -1,0 +1,236 @@
+---
+keyword: "additionalProperties"
+signature: "Schema"
+summary: "Validation succeeds if the schema validates against each value not matched by other object applicators in this vocabulary."
+kind: [ "applicator", "annotation" ]
+instance: [ "object" ]
+specification: "https://json-schema.org/draft/2020-12/json-schema-core.html#section-10.3.2.3"
+metaschema: "https://json-schema.org/draft/2020-12/meta/applicator"
+index: -998
+introduced_in: draft1
+interdependencies:
+  - vocabulary: applicator
+    keyword: properties
+  - vocabulary: applicator
+    keyword: patternProperties
+related:
+  - vocabulary: applicator
+    keyword: dependentSchemas
+  - vocabulary: applicator
+    keyword: propertyNames
+  - vocabulary: validation
+    keyword: required
+  - vocabulary: validation
+    keyword: dependentRequired
+  - vocabulary: validation
+    keyword: minProperties
+  - vocabulary: validation
+    keyword: maxProperties
+  - vocabulary: unevaluated
+    keyword: unevaluatedProperties
+---
+
+Annotations
+-----------
+
+The annotation result of this keyword is the set of instance property names validated by this keyword.
+
+## Explanation
+
+The `additionalProperties` keyword is used to control the handling of properties whose names are not listed in the `properties` keyword or match any of the regular expressions in the `patternProperties` keyword. By default any additional properties are allowed.
+
+The behavior of this keyword depends on the presence and annotation results of `properties` and `patternProperties` within the same schema object. Validation with `additionalProperties` applies only to the child values of instance names that do not appear in the annotation results of either `properties` or `patternProperties`.
+
+* The value of `additionalProperties` must be a valid JSON Schema.
+* Each property value of this object must be a valid JSON Schema.
+* The annotation result of this keyword is the set of instance property names validated by this keyword's subschema.
+* This annotation affects the behavior of `unevaluatedProperties` in the Unevaluated vocabulary.
+* Omitting this keyword has the same assertion behavior as an empty object.
+
+## Examples
+
+{{<schema `Schema with 'additionalProperties' set to boolean false`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "properties": {
+    "foo": { "type": "string" }
+  },
+  "additionalProperties": false
+}
+{{</schema>}}
+
+{{<instance-pass `An instance with no additional properties is valid`>}}
+{ "foo": "foo" }
+{{</instance-pass>}}
+
+{{<instance-annotation>}}
+[
+  // ...
+  {
+    "valid": true,
+    "keywordLocation": "/properties",
+    "instanceLocation": "",
+    "annotation": [ "foo" ]
+  },
+  // ...
+]
+{{</instance-annotation>}}
+
+{{<instance-fail `An instance with additional properties is invalid`>}}
+{ "foo": "foo", "bar": "bar" }
+{{</instance-fail>}}
+* _When `additionalProperties` is set to false, all the instance properties must either be present in the `properties` or match any regex within `patternProperties`; otherwise, the validaion will fail._
+
+{{<schema `Schema with 'additionalProperties' set to an object schema`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "properties": {
+    "name": { "type": "string" }
+  },
+  "additionalProperties": {
+    "type": "number"
+  }
+}
+{{</schema>}}
+
+{{<instance-pass `An object instance with properties conforming to the schema is valid`>}}
+{ "name": "John Doe", "age": 21 }
+{{</instance-pass>}}
+
+{{<instance-annotation>}}
+[
+  // ...
+  {
+    "valid": true,
+    "keywordLocation": "/properties",
+    "instanceLocation": "",
+    "annotation": [ "name" ]
+  },
+  {
+    "valid": true,
+    "keywordLocation": "/additionalProperties",
+    "instanceLocation": "",
+    "annotation": [ "age" ]
+  },
+  // ...
+]
+{{</instance-annotation>}}
+
+{{<instance-fail `The value of 'age' must be a number`>}}
+{ "name": "John Doe", "age": "21" }
+{{</instance-fail>}}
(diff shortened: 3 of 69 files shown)


Era-cell:
I have updated and implemented:
1. scroll of sidebar and main content function separately - I strongly felt that's neccesary
2. Addressing the issue - https://github.com/Intelligence-AI/learnjsonschema.com/issues/36 I have included alphabetical ordering or sidebar
3. And switching between vocabulary and alphabetical ordering - 
      - vocabularies implementing the dropdown , !!! having split button left lists down all keywords right redirects to vocabulary 
      -     page. this should be considered and implemented with care with highlights on users ease of understanding and usage.
      
4. Switch between dialects needs styling...
5. At the end - -- And do we need to implement ajax to keep the sidebar state when redirected

AgniveshChaubey:
Just some UI suggestions...

image

Maybe keeping the radio slider without arrows would be more visually appealing?
Also what about making the dialect dropdown cover the entire width of the sidebar (with some horizontal margins)?

image
There's an unnecessary scrollbar at the end of the sidebar. 

> 5. At the end - -- And do we need to implement ajax to keep the sidebar state when redirected

image

The collapse and open state of accordions are already maintained upon redirection, so maybe it's also possible to keep the state of the sidebar without any additional library?

Era-cell:
> Just some UI suggestions...
> 
> image
> 
> Maybe keeping the radio slider without arrows would be more visually appealing? Also what about making the dialect dropdown cover the entire width of the sidebar (with some horizontal margins)?
> 
> image There's an unnecessary scrollbar at the end of the sidebar.
> 
> > 5. At the end - -- And do we need to implement ajax to keep the sidebar state when redirected
> 
> image
> 
> The collapse and open state of accordions are already maintained upon redirection, so maybe it's also possible to keep the state of the sidebar without any additional library?

Cool!! I have written my own logic for accordion javascript.
Can you tell me how to make the javascript in bootstrap work for our project, its css works but interactions werent working...
I juts cloned the project and used "hugo server" did I had to do something for bootstrap to work completely

Era-cell:
And I think just the radio is enough and we can keep it single colored right, no need of grey and blue.. I guess

Era-cell:
> > And I think just the radio is enough and we can keep it single colored right, no need of grey and blue.. I guess
> 
> Yes, I guess that make sense because changing the color somewhat represents enabling/disabling something. But here we are just switching the list type.

Exactly, and I am not sure what words should be used for "alphabetical ordering" and group by vocabularies
"group by vocabularies" will be too long, ..

jviotti:
The UI is getting nice, but probably quite a bunch to polish before we can merge. For example, I'd really like to reduce JavaScript use to a minimum, avoid some often discouraged CSS features like `calc`, inline CSS styles, etc.

What I recommend is splitting this PR into two, as there are two things we are trying to cover here:

- First one is allowing collapsing vocabularies / dialects in the sidebar. Let's do that in one focused PR. I even believe you can collapse with just CSS + HTML
- Second one is a dropdown to switch between dialects. I think that one too you can do it with much less JavaScript, or even none at all. For example, what if we make the static HTML at `/2020-12` only show the 2020-12 dialect in the sidebar and `/2019-09` only show the 2019-09 dialect in the sidebar? If so, the dropdown could just perform a page redirection, without any JavaScript?

jviotti:
Now that I write that last comment, I wonder if we even need the collapsing feature with the approach I mentioned on the second part... Maybe we just re-direct and show all of the vocabularies on the currently selected dialect?

Era-cell:
> The UI is getting nice, but probably quite a bunch to polish before we can merge. For example, I'd really like to reduce JavaScript use to a minimum, avoid some often discouraged CSS features like `calc`, inline CSS styles, etc.
> 
> What I recommend is splitting this PR into two, as there are two things we are trying to cover here:
> 
> * First one is allowing collapsing vocabularies / dialects in the sidebar. Let's do that in one focused PR. I even believe you can collapse with just CSS + HTML
> * Second one is a dropdown to switch between dialects. I think that one too you can do it with much less JavaScript, or even none at all. For example, what if we make the static HTML at `/2020-12` only show the 2020-12 dialect in the sidebar and `/2019-09` only show the 2019-09 dialect in the sidebar? If so, the dropdown could just perform a page redirection, without any JavaScript?

For 1st one: like: https://www.websiterating.com/resources/html-only-accordion-without-js-css/#:~:text=The%20%3Cdetails%3E%20and%20%3Csummary%3E%20HTML5%20tags%20are%20officially,accordion.%20Meaning%20no%20JS%20OR%20CSS%20is%20needed i'll try it

2nd, I tried to use bootstraps dropdown, 
!! yeah can you tell me how to use the javascript provided by bootstrap, is it by including what all files we need into main.js file 
in assets. We can use javascript provided by bootstrap right? That didn't work in my local ...

Era-cell:
> Now that I write that last comment, I wonder if we even need the collapsing feature with the approach I mentioned on the second part... Maybe we just re-direct and show all of the vocabularies on the currently selected dialect?

Ok.. this means we will just show all the vocabualries all closed, and user can open whichever he wants from that point..
Yeah!! because there may and maynot be some vocabularies like format assertion and annotation in 2020-12 isnt present in 2019-09 so we cannot keep that open..

jviotti:
> this means we will just show all the vocabualries all closed,

The opposite. They would just show all as expanded, exactly as they are right now on the live site.

> and user can open whichever he wants from that point..

This is correct. The user can swap dialects using the dropdown.

> because there may and maynot be some vocabularies like format assertion and annotation in 2020-12 isnt present in 2019-09 so we cannot keep that open..

You should never encounter that problem. The contents of the sidebar are not hardcoded, but auto-generated given the folders in the content. i.e. https://github.com/Intelligence-AI/learnjsonschema.com/blob/main/layouts/partials/sidebar.html#L8.

If your 2019-09 folder didn't have vocabulary X, then when visiting the landing page for 2019-09 (which would show the 2019-09 sidebar), then vocabulary X would never be generated there.

jviotti:
At least that ^ feels to me like a doable and simple solution to at least start hosting more dialects, and it would require minimal JavaScript, if any.

Era-cell:
@jviotti.., I can use bootstrap for javascript right??
I tried many methods to make the javascript of the bootstrap work(right now only css from bootstrap is working and not JS):
1. using CDN JS works partially
2. using bootstrap.min.js also works partially
3. importing in assets/main.js
Like :

```
// import Dropdown from '../vendor/bootstrap/js/src/dropdown.js';
import Tooltip from './vendor/bootstrap/js/src/tooltip.js';

// Initialize tooltips
for (const element of document.querySelectorAll('[data-bs-toggle="tooltip"]')) {
  new Tooltip(element);
}

// for (const element of document.querySelectorAll('.dropdown')) {
//   new Dropdown(element);
// }

```
as its done for tooltip.js similarly I have included dropdown.js but it gives this error:
` Could not resolve "../vendor/bootstrap/js/src/dropdown.js"`
So, tell me how can I use the javascript from bootstrap if I can use it..

Era-cell:
If javascript logic from bootstrap works, yeh! we will have hardly few lines of js or no at all

jviotti:
> If javascript logic from bootstrap works, yeh! we will have hardly few lines of js or no at all, ...

Well, we wouldn't be writing JS ourselves, but we would be running plenty of JS from Bootstrap. That is better than writing a lot of custom JS, but if we can do it without any JS, then even better?

What do you want Bootstrap JS for?

jviotti:
In any case, independently of whether we do Bootstrap JS or not, the "correct" way of doing it would be out of  https://github.com/Intelligence-AI/learnjsonschema.com/tree/main/vendor/bootstrap/js/src and Hugo itself, as it supports bundling JavaScript, etc: https://gohugo.io/hugo-pipes/js/

jviotti:
But I believe we can pull this off without any JavaScript at all, and that'd be ideal. Much faster, less for the browser to download & evaluate, etc

jviotti:
Ah, I just realized we were using a bit of Bootstrap JavaScript already as you pointed out above: https://github.com/Intelligence-AI/learnjsonschema.com/blob/main/assets/main.js

jviotti:
Your paths are wrong:

```js
// import Dropdown from '../vendor/bootstrap/js/src/dropdown.js';
import Tooltip from './vendor/bootstrap/js/src/tooltip.js';
```

The first one has `../vendor` while the second one has `./vendor`

Era-cell:
1st one you mean by using <details> and summary tags right

> > If javascript logic from bootstrap works, yeh! we will have hardly few lines of js or no at all, ...
> 
> Well, we wouldn't be writing JS ourselves, but we would be running plenty of JS from Bootstrap. That is better than writing a lot of custom JS, but if we can do it without any JS, then even better?
> 
> What do you want Bootstrap JS for?

I needed it for accordion functionality and switch. Like switch between alphabetical and group by vocab. 
If accordion can be implemented using <details><summary> I have to see whether <details>summary method can have anchors and dropdown both functionalities inside them.

jviotti:
Let's figure out the dropdown first, to swap between dialects while showing a sidebar that is exactly what we have now. I think once that's done, we can turn our attention to the accordion, potentially using Bootstrap, and potential grouping. But I think the former is the most important change

jviotti:
@Era-cell So there are two problems we are trying to solve:

1. Allowing to navigate to dialects other than 2020-12, showing different sidebar content depending on the dialect you are visiting
2. Implementing accordion functionality, etc to improve the UX of the sidebar with the content they already have

For number 2, I agree we probably need JavaScript to do it well, and we can likely use Bootstrap's accordion plugins, etc. However, keep in mind that in importance, task 1 is much more important than task 2, so I propose we park task 2 for a little bit until we properly solve task 1.

So focusing on task 1, the flow I'm imagining, hopefully without involving JavaScript at all (but correct me if I'm wrong, as I'm not a web developer) is:

- Imagine you are at `learnjsonschema.com/2020-12`. Because you are visiting 2020-12, the left sidebar shows you 2020-12 vocabularies and keywords, exactly like what you see on the live site right now
- On the sidebar, you see a dropdown to change dialects
- You click on the dropdown to see the relevant options
- When you select `2019-09` from the dropdown, we will perform a full page redirection to `learnjsonschema.com/2019-09`
- When `learnjsonschema.com/2019-09` loads, its sidebar contains vocabularies and keywords only for 2019-09, all expanded

Era-cell:
I'll just show you difference between them using recordings:
this one can be achieved, we can redirect successfully, uses only dropdown

https://github.com/Intelligence-AI/learnjsonschema.com/assets/77973415/58878b5b-8f11-48ae-9484-0e63b6506cd9

2. using only select:
No redirection but selected will be displayed
https://github.com/Intelligence-AI/learnjsonschema.com/assets/77973415/c37be471-f516-4eb5-9f60-9f208b3dc2e8

3. using javascript and select:

https://github.com/Intelligence-AI/learnjsonschema.com/assets/77973415/586bded6-7c02-4790-aefc-623ae98fc323

4. similarly if dropdown we have to write javascript to show the selected one, like select shows..

jviotti:
Assuming I get it correctly (hard to tell without seeing the code), what seems like the best option to me is:

- Using Bootstrap's Dropdown component (https://getbootstrap.com/docs/5.3/components/dropdowns), which if it uses JavaScript, I guess that's OK
- Make that component redirect with `<a>`  when selecting an option

Era-cell:
> Assuming I get it correctly (hard to tell without seeing the code), what seems like the best option to me is:
> 
> * Using Bootstrap's Dropdown component (https://getbootstrap.com/docs/5.3/components/dropdowns), which if it uses JavaScript, I guess that's OK
> * Make that component redirect with `<a>`  when selecting an option

So, this is option1... It redirects, but keeping the button saying "Select dialect" will be necessary

jviotti:
> So, this is option1... It redirects, but keeping the button saying "Select dialect" will be necessary

I think we can handle this too using Hugo templates. For example, we can set the current dialect as the title of the dropdown, and omit the option for the same dialect on the option list.

Era-cell:
> > So, this is option1... It redirects, but keeping the button saying "Select dialect" will be necessary
> 
> I think we can handle this too using Hugo templates. For example, we can set the current dialect as the title of the dropdown, and omit the option for the same dialect on the option list.

Yeah, obviously that's handled by Hugo templates, you don't worry about that. It will be extensible.
The main problem is what you said Current Dialect has to be set as Title of the dropdown, to achieve that we need javascript.
That isn't provided by dropdowns - dropdowns are meant to have the same title always... I guess you get ..😕

jviotti:
> The main problem is what you said Current Dialect has to be set as Title of the dropdown, to achieve that we need javascript.
That isn't provided by dropdowns - dropdowns are meant to have the same title always

The title of the dropdown is determined by the string we have in the `<button>` tag, and that we can still interpolate using Hugo, without JavaScript.

Try to send a PR implementing the option 1 we discussed, and I can point out in code what I mean.

Era-cell:
> > The main problem is what you said Current Dialect has to be set as Title of the dropdown, to achieve that we need javascript.
> > That isn't provided by dropdowns - dropdowns are meant to have the same title always
> 
> The title of the dropdown is determined by the string we have in the `<button>` tag, and that we can still interpolate using Hugo, without JavaScript.
> 
> Try to send a PR implementing the option 1 we discussed, and I can point out in code what I mean.

Oh!! I didn't think this way. I got it now, we can also render selected dialect using hugo

jviotti:
Exactly! :)

Era-cell:
> Exactly! :)

I was using selected_dialect to display if statement, but for this one I got dumb

PR #165, opened 2024-05-03
Adding dropdown,

Era-cell:
Features added: dropdown on hover
Which redirects to syntax http://base.com/{selected_dialect}

### DIFF
--- a/assets/main.scss
+++ b/assets/main.scss
@@ -74,8 +74,7 @@ $bootstrap-icons-font-dir: "./";
 }
 
 #sidebar {
-  // min-width: 265px;
-  overflow-y: auto
+  min-width: 260px;
 }
 
 #sidebar > div {
@@ -129,8 +128,4 @@ article table {
 
 .alert > :last-child {
   @extend .mb-0;
-}
-
-#dialect-dropdown{
-  padding-right: 10px;
 }
--- a/assets/main.scss
+++ b/assets/main.scss
@@ -24,7 +24,7 @@ $warning: #ffff6a;
 @import "../vendor/bootstrap/scss/forms";
 @import "../vendor/bootstrap/scss/buttons";
 // @import "../vendor/bootstrap/scss/transitions";
-// @import "../vendor/bootstrap/scss/dropdown";
+@import "../vendor/bootstrap/scss/dropdown";
 // @import "../vendor/bootstrap/scss/button-group";
 @import "../vendor/bootstrap/scss/nav";
 @import "../vendor/bootstrap/scss/navbar";
--- a/assets/main.scss
+++ b/assets/main.scss
@@ -74,7 +74,8 @@ $bootstrap-icons-font-dir: "./";
 }
 
 #sidebar {
-  min-width: 265px;
+  // min-width: 265px;
+  overflow-y: auto
 }
 
 #sidebar > div {
@@ -82,7 +83,6 @@ $bootstrap-icons-font-dir: "./";
   @extend .top-0;
   @extend .start-0;
   @extend .bottom-0;
-  overflow-y: scroll;
 }
 
 // Tables with vertical headers collapse the headers by default
@@ -129,4 +129,8 @@ article table {
 
 .alert > :last-child {
   @extend .mb-0;
+}
+
+#dialect-dropdown{
+  padding-right: 10px;
 }
--- a/layouts/_default/section.html
+++ b/layouts/_default/section.html
@@ -6,22 +6,25 @@
 {{ $dialectURL := $dialect | printf "/%s" }}
 {{ $githubURL := .File.Path | printf "%s/blob/main/content/%s" .Site.Params.GitHub }}
 
+{{ $siteSections := .Site.Sections }}
 {{ if eq .Page.Section .Title }}
   {{ partial "breadcrumb.html" (dict
     "dialect" $dialect
-    "githubURL" $githubURL ) }}
+    "githubURL" $githubURL
+    "siteSections" $siteSections ) }}
 {{ else }}
   {{ $vocabulary := .Title }}
   {{ partial "breadcrumb.html" (dict
     "dialect" $dialect
     "dialectURL" (ref . $dialectURL)
     "vocabulary" $vocabulary
-    "githubURL" $githubURL ) }}
+    "githubURL" $githubURL
+    "siteSections" $siteSections ) }}
 {{ end }}
 
 <div class="d-flex bg-body flex-fill">
   {{ partial "sidebar.html" . }}
-  <div class="flex-xl-grow-1 mw-100">
+  <div id="test-100"  class="flex-xl-grow-1 mw-100">
     <main class="px-3 px-lg-5 pt-4 pb-3">
       <header class="mb-3">
         <h1>{{ .Title }}
--- a/layouts/_default/section.html
+++ b/layouts/_default/section.html
@@ -24,7 +24,7 @@
 
 <div class="d-flex bg-body flex-fill">
   {{ partial "sidebar.html" . }}
-  <div id="test-100"  class="flex-xl-grow-1 mw-100">
+  <div class="flex-xl-grow-1 mw-100">
     <main class="px-3 px-lg-5 pt-4 pb-3">
       <header class="mb-3">
         <h1>{{ .Title }}
--- a/layouts/partials/breadcrumb.html
+++ b/layouts/partials/breadcrumb.html
@@ -8,10 +8,10 @@
     <ul class="dropdown-menu">
       {{ range .siteSections }}
         {{ if ne $selected_dialect .Title }}
-          <li><a class="dropdown-item disabled" aria-disabled="true" href="{{ ref . .RelPermalink }}">{{ .Title }}</a></li>
+          <li><a class="dropdown-item" href="{{ ref . .RelPermalink }}">{{ .Title }}</a></li>
         {{ end }}
       {{ end }}
-      <li><a class="dropdown-item"> More coming soon! </a></li>
+      <li><a class="dropdown-item disabled" aria-disabled="true">More coming soon!</a></li>
     </ul>
   </div>
 
--- a/layouts/partials/breadcrumb.html
+++ b/layouts/partials/breadcrumb.html
@@ -1,13 +1,19 @@
 <nav aria-label="breadcrumb" class="d-flex align-items-center bg-light bg-gradient border-bottom px-3 py-1 shadow-sm">
 
-  <div id="dialect-dropdown" class="dropdown mt-1">
-    <button class="btn btn-outline-primary dropdown-toggle" type="button" data-bs-toggle="dropdown" aria-expanded="false">
+  {{$selected_dialect := .dialect}}
+  <div class="dropdown mt-1 pe-3" >
+    <button class="btn btn-outline-primary dropdown-toggle btn-sm" type="button" data-bs-toggle="dropdown" aria-expanded="false">
       Other dialects
     </button>
-    <ul class="dropdown-menu">
+    <!-- <ul class="dropdown-menu">
       {{ range .siteSections }}
+        {{if ne $selected_dialect .Title}}
           <li><a class="dropdown-item" href="{{ ref . .RelPermalink }}">{{ .Title }}</a></li>
+        {{end}}
       {{ end }}
+    </ul> -->
+    <ul class="dropdown-menu">
+      <li><a class="dropdown-item fs-6">More dialects coming soon!</a></li>
     </ul>
   </div>
 
--- a/layouts/partials/breadcrumb.html
+++ b/layouts/partials/breadcrumb.html
@@ -1,19 +1,17 @@
 <nav aria-label="breadcrumb" class="d-flex align-items-center bg-light bg-gradient border-bottom px-3 py-1 shadow-sm">
 
-  {{$selected_dialect := .dialect}}
-  <div class="dropdown mt-1 pe-3" >
+  {{ $selected_dialect := .dialect }}
+  <div class="dropdown pe-3">
     <button class="btn btn-outline-primary dropdown-toggle btn-sm" type="button" data-bs-toggle="dropdown" aria-expanded="false">
       Other dialects
     </button>
-    <!-- <ul class="dropdown-menu">
-      {{ range .siteSections }}
-        {{if ne $selected_dialect .Title}}
-          <li><a class="dropdown-item" href="{{ ref . .RelPermalink }}">{{ .Title }}</a></li>
-        {{end}}
-      {{ end }}
-    </ul> -->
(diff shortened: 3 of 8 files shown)


layouts/partials/sidebar.html
```diff
@@ -1,9 +1,23 @@
 <aside id="sidebar" class="border-end position-relative d-none d-xl-block">
   <div class="px-4 py-4">
+
+    {{$selected_dialect := index (split .RelPermalink "/") 1}}
+    <div class="dropdown">
+      <button class="dropdown-button fw-bold ps-0" type="button" id="dropdownMenuButton" data-bs-toggle="dropdown" aria-expanded="false">
```
jviotti, line 6: Can we style this button a more prominently and differently than the vocabulary accordions? Right now I feel like its not clear that its even clickable

jviotti: I also noticed that this dropdown opens automatically when you however, while the Bootstrap one gets opened only when you click it? Can we get Bootstrap click behavior instead?

layouts/partials/sidebar.html
```diff
@@ -38,3 +52,57 @@
     <small class="text-muted">More dialects coming soon!</small>
   </div>
 </aside>
+
+<style>
```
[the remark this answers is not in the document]
jviotti: In general, the point of Bootstrap is that they have a lot of pre-made components that are even customizable using SCSS variables so that you can avoid writing actual CSS in the vast majority of cases

layouts/partials/sidebar.html
Era-cell: Wait.. can I get help on this: for the javascript to 

Uploading Recording 2024-05-07 004034.mp4…

work I am assigning objects as you have done in the tooltip like this,
Screenshot 2024-05-07 003740
But seems that javascript from bootstrap isn't working properly or it behaving this way by the affect of hugo components,

jviotti:
@Era-cell There are a few things missing, some of which were non-obvious:

- You need to enable the dropdown SCSS partials from Bootstrap in `assets/main.scss`
- The HTML for the dropdown was incorrect. For example, the button should have had the class `dropdown-toggle` instead of `dropdown-button` and the anchors needed the class `dropdown-item` (which the documentation shows)
- You had to tweak `vendor/bootstrap.mask` and re-run [`vendorpull`](https://github.com/sourcemeta/vendorpull) to actually pull the Bootstrap dropdown JavaScript plugin
- Finally, instantiate the plugin in `assets/main.js`
- Plus you can get rid of the inline styles you had there

Let me do at least some of those changes for you and push a commit here

jviotti:
Done. Give it a shot yourself. Though now that I play with it, I wonder if putting the dropdown on the sidebar is the right place... maybe same dropdown but on either the breadcrumb or the navigation bar would feel better... Let me play with it a bit

jviotti:
I went back and forth on it a lot, but I think I quite like it in the breadcrumb, and instead of saying "2020-12" we say something like "Other dialects" or "Explore other dialects":

Screenshot 2024-05-06 195805

What do you think? I did with a `btn btn-sm btn-outline-primary`

Era-cell:
Oh!!  now I get it. I was masked in bootstrap.mask, 😵‍💫 I wasted a lot of time

Era-cell:
> I went back and forth on it a lot, but I think I quite like it in the breadcrumb, and instead of saying "2020-12" we say something like "Other dialects" or "Explore other dialects":
> 
> Screenshot 2024-05-06 195805
> 
> What do you think? I did with a `btn btn-sm btn-outline-primary`

That looks more better, even I asked others. They too felt this intuitive rather than having it inside sidebar

Era-cell:
@jviotti I have moved the dropdown to breadcrumb, maybe this time its good..

layouts/partials/sidebar.html
```diff
@@ -1,9 +1,11 @@
-<aside id="sidebar" class="border-end position-relative d-none d-xl-block">
-  <div class="px-4 py-4">
+<aside class="border-end position-relative d-none d-xl-block">
+  <div class="px-4 py-2">
+
+    {{$selected_dialect := index (split .RelPermalink "/") 1}}
```
jviotti, line 4: @Era-cell This variable is completely unused. If so, please delete it

layouts/partials/sidebar.html
```diff
@@ -1,9 +1,11 @@
-<aside id="sidebar" class="border-end position-relative d-none d-xl-block">
```
jviotti, line 1: Why was the `id` removed?

layouts/_default/section.html
```diff
@@ -6,22 +6,25 @@
...
     "vocabulary" $vocabulary
-    "githubURL" $githubURL ) }}
+    "githubURL" $githubURL
+    "siteSections" $siteSections ) }}
 {{ end }}
 
 <div class="d-flex bg-body flex-fill">
   {{ partial "sidebar.html" . }}
-  <div class="flex-xl-grow-1 mw-100">
+  <div id="test-100"  class="flex-xl-grow-1 mw-100">
```
jviotti, line 27: We shouldn't be adding any "test" `id`s, even more if they will just go unused.

layouts/partials/breadcrumb.html
```diff
@@ -1,4 +1,16 @@
-<nav aria-label="breadcrumb" class="d-flex align-items-center bg-light bg-gradient border-bottom px-3 py-2 shadow-sm">
+<nav aria-label="breadcrumb" class="d-flex align-items-center bg-light bg-gradient border-bottom px-3 py-1 shadow-sm">
+
+  <div id="dialect-dropdown" class="dropdown mt-1">
```
jviotti, line 3: `id`s should be added very sparingly. I don't think you need it here.

assets/main.scss
```diff
@@ -130,3 +130,7 @@ article table {
 .alert > :last-child {
   @extend .mb-0;
 }
+
+#dialect-dropdown{
+  padding-right: 10px;
```
jviotti, line 135: Please no ad-hoc CSS styles when Bootstrap has pre-defined class helpers. In this case, you can use the `pe-X` padding helpers directly on the HTML. Take a look at the documentation for how to set padding in Bootstrap

assets/main.scss
```diff
@@ -74,15 +74,15 @@ $bootstrap-icons-font-dir: "./";
 }
 
 #sidebar {
-  min-width: 265px;
+  // min-width: 265px;
```
jviotti, line 77: You can delete this commented style.

assets/main.scss
```diff
@@ -74,15 +74,15 @@ $bootstrap-icons-font-dir: "./";
...
+  overflow-y: auto
```
jviotti, line 78: What is the reasoning for these `overflow` changes? If they improve the sidebar (I'm no longer seeing a scrolling bar, which is cool!) then let's send it as a separate PR and merge that first?

layouts/partials/breadcrumb.html
```diff
@@ -1,4 +1,16 @@
...
+    <button class="btn btn-outline-primary dropdown-toggle" type="button" data-bs-toggle="dropdown" aria-expanded="false">
```
jviotti, line 4: Can you do a small button with `btn-sm` here? The current looks too big IMO

layouts/partials/breadcrumb.html
```diff
@@ -1,4 +1,16 @@
...
+      Other dialects
+    </button>
+    <ul class="dropdown-menu">
+      {{ range .siteSections }}
+          <li><a class="dropdown-item" href="{{ ref . .RelPermalink }}">{{ .Title }}</a></li>
```
jviotti, line 9: Can you tweak the dropdown options to not display the current dialect? i.e. if you are on 2020-12 and you open the dropdown, you shouldn't see 2020-12. 

Because we don't have any other dialect right now, that means that the dropdown would be empty. What we can do is show a disabled dropdown entry that says "More dialects coming soon!" or something like that, like we originally had on the sidebar?

jviotti (changes requested):
I left a few comments, but shaping up, and it does work already

layouts/partials/sidebar.html
Era-cell: 😅 my bad, I badly need to review my PR before asking for review. I'll note it

layouts/partials/breadcrumb.html
Era-cell: Okay.. I am now using bootstrap classes

assets/main.scss
Era-cell: Yeah! I  wanted the sidebar bar to scrollabe. Lets do it another PR

layouts/partials/breadcrumb.html
Era-cell: Okay.. what I have done is commented out the options, because however we will do it ASAP..

layouts/partials/breadcrumb.html
Era-cell: I did change the button to small, but what about the dropdown elements? The minimum text sizing we can give using bootstrap class I guess is "fs-6" which is same as the default size.

layouts/partials/breadcrumb.html
```diff
@@ -1,4 +1,22 @@
...
+    </button>
+    <!-- <ul class="dropdown-menu">
+      {{ range .siteSections }}
+        {{if ne $selected_dialect .Title}}
+          <li><a class="dropdown-item" href="{{ ref . .RelPermalink }}">{{ .Title }}</a></li>
+        {{end}}
+      {{ end }}
+    </ul> -->
+    <ul class="dropdown-menu">
+      <li><a class="dropdown-item fs-6">More dialects coming soon!</a></li>
```
jviotti, line 16: Please make this a disabled item: https://getbootstrap.com/docs/5.3/components/dropdowns/#disabled

jviotti: Also now that I see it, this feels a bit long. Maybe just "More coming soon!"?

layouts/partials/breadcrumb.html
```diff
@@ -1,4 +1,22 @@
...
+  {{$selected_dialect := .dialect}}
+  <div class="dropdown mt-1 pe-3" >
+    <button class="btn btn-outline-primary dropdown-toggle btn-sm" type="button" data-bs-toggle="dropdown" aria-expanded="false">
+      Other dialects
+    </button>
+    <!-- <ul class="dropdown-menu">
+      {{ range .siteSections }}
+        {{if ne $selected_dialect .Title}}
+          <li><a class="dropdown-item" href="{{ ref . .RelPermalink }}">{{ .Title }}</a></li>
+        {{end}}
```
jviotti, line 12: There is a lot of space inconsistencies in this snippet. For example, here you say `{{end}}`, but below you say `{{ end }}`. Also above you say `{{$selected_dialect := .dialect}}` where it should have been `{{ $selected_dialect := .dialect }}`. Make sure style is consistent with the rest of the project

layouts/partials/breadcrumb.html
```diff
@@ -1,4 +1,22 @@
-<nav aria-label="breadcrumb" class="d-flex align-items-center bg-light bg-gradient border-bottom px-3 py-2 shadow-sm">
+<nav aria-label="breadcrumb" class="d-flex align-items-center bg-light bg-gradient border-bottom px-3 py-1 shadow-sm">
+
+  {{$selected_dialect := .dialect}}
+  <div class="dropdown mt-1 pe-3" >
```
jviotti, line 4: There is a weird space here before `>`. Be mindful of those details!

jviotti: Also, the `mt-1` is ruining the vertical alignment. I think you should get rid of that class

Era-cell: I have removed it...

layouts/partials/breadcrumb.html
```diff
@@ -1,4 +1,20 @@
...
+<nav aria-label="breadcrumb" class="d-flex align-items-center bg-light bg-gradient border-bottom px-3 py-1 shadow-sm">
+
+  {{ $selected_dialect := .dialect }}
+  <div class="dropdown pe-3">
+    <button class="btn btn-outline-primary dropdown-toggle btn-sm" type="button" data-bs-toggle="dropdown" aria-expanded="false">
+      Other dialects
+    </button>
+    <ul class="dropdown-menu">
+      {{ range .siteSections }}
+        {{ if ne $selected_dialect .Title }}
```
jviotti, line 10: This doesn't look right. You are looping over all sections (which there are none?) and making them disabled, while still adding the option for "More coming soon!" below without making that disabled (which is the one that should have been disabled)

layouts/partials/breadcrumb.html
```diff
@@ -1,4 +1,20 @@
...
+          <li><a class="dropdown-item disabled" aria-disabled="true" href="{{ ref . .RelPermalink }}">{{ .Title }}</a></li>
+        {{ end }}
+      {{ end }}
+      <li><a class="dropdown-item"> More coming soon! </a></li>
```
jviotti, line 14: Also, you added spaces before/after the square brackets, which is not the case elsewhere:

```suggestion
      <li><a class="dropdown-item">More coming soon!</a></li>
```

layouts/partials/breadcrumb.html
Era-cell: Should I disable "More coming soon!" ? But right now we have only 2020-12. So, shouldn't that be enabled rn. I didn't understand the need of "More coming soon!" if its always disabled. If we enable all sections, right now the dropdown will display nothing.. which may make the user feel, why is the "Other dialect" button even existing.

jviotti: > Should I disable "More coming soon!" ?

Correct. That doesn't make sense as clickable, as where would it redirect if so?

>  But right now we have only 2020-12. So, shouldn't that be enabled rn

No. Because we only support 2020-12, you are always within the 2020-12 context, so displaying 2020-12 in the dropdown again doesn't make sense. The dropdown should list the options that are NOT the current one. There is no point in making the user navigate to the same place they are already in!

> If we enable all sections, right now the dropdown will display nothing.. which may make the user feel, why is the "Other dialect" button even existing.

The dropdown wouldn't be empty. It'd display a single disabled element: "More coming soon!"

Era-cell: > No. Because we only support 2020-12, you are always within the 2020-12 context, so displaying 2020-12 in the dropdown again doesn't make sense. The dropdown should list the options that are NOT the current one. There is no point in making the user navigate to the same place they are already in!

Oh shet! "that" I mean the loop. However I understood the importance of disabled item. I guess I have got it right this time

layouts/partials/breadcrumb.html
Era-cell: @jviotti I have made the changes.

PR #167, opened 2024-05-06
Keyword `unevaluatedItems`

AgniveshChaubey:
This PR addresses #84.

### DIFF
--- a/content/2020-12/unevaluated/unevaluatedItems.markdown
+++ b/content/2020-12/unevaluated/unevaluatedItems.markdown
@@ -34,9 +34,11 @@ If this keyword is applied to any instance element, it produces an annotation va
 
 ## Explanation
 
-First, let's understand what evaluation means. `unevaluatedItems` considers annotations from `prefixItems`, `items`, and `contains`, both as adjacent keywords and in subschemas of adjacent keywords. Additionally, `unevaluatedItems` is affected by other `unevaluatedItems` in nested schemas (if present). Each of these keywords will produce an annotation of the locations that they've evaluated in the array.
+First, let's understand what evaluation means. `unevaluatedItems` considers annotations from `prefixItems`, `items`, and `contains`, both as adjacent keywords and in subschemas of adjacent keywords. Additionally, `unevaluatedItems` is affected by other `unevaluatedItems` in nested schemas (if present). Each of these keywords will produce an annotation of the indexes that they've evaluated in the array.
 - For `items`, it's boolean true.
 - For `prefixItems`, it is the largest index to which this keyword applied a subschema (the value may be a boolean true if a subschema was applied to every index of the instance).
 - For `contains`, it is an array of the indexes to which this keyword validates successfully when applying its subschema, in ascending order (the value may be a boolean true if the subschema validates successfully when applied to every index of the instance).
 
 If any of these are in subschemas of adjacent keywords, and those subschemas fail validation, those annotations are dropped in that case. The effect is that those properties are not considered evaluated.
+
+Validation with `unevaluatedItems` applies only to the indexes of the array instance that do not appear in the `properties`, `patternProperties`, `additionalProperties`, or `unevaluatedProperties` annotation results that apply to the instance location being validated.
--- a/content/2020-12/unevaluated/unevaluatedItems.markdown
+++ b/content/2020-12/unevaluated/unevaluatedItems.markdown
@@ -42,3 +42,333 @@ First, let's understand what evaluation means. `unevaluatedItems` considers anno
 If any of these are in subschemas of adjacent keywords, and those subschemas fail validation, those annotations are dropped in that case. The effect is that those properties are not considered evaluated.
 
 Validation with `unevaluatedItems` applies only to the indexes of the array instance that do not appear in the `properties`, `patternProperties`, `additionalProperties`, or `unevaluatedProperties` annotation results that apply to the instance location being validated.
+
+## Examples
+
+{{<schema `Schema with 'unevaluatedItems' set to boolean true`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "unevaluatedItems ": true
+}
+{{</schema>}}
+
+{{<instance-pass `All object instances pass against the true schema`>}}
+[ "foo": "bar" ]
+{{</instance-pass>}}
+
+{{<instance-pass `'unevaluatedItems' does not have any effect on instances other than an array`>}}
+"John Doe"
+{{</instance-pass>}}
+
+{{<instance-annotation>}}
+[
+  // ...
+  {
+    "valid": true,
+    "keywordLocation": "/unevaluatedItems",
+    "instanceLocation": "",
+    "annotation": "true"
+  },
+  // ...
+]
+{{</instance-annotation>}}
+* Here, no items are defined in the above schema. Consequently, all items in an array instance are considered unevaluated, and the `unevaluatedItems` subschema applies to them. Since the subschema here is a boolean true, an instance with unevaluated items, regardless of their value, is considered valid.
+
+{{<schema `Schema with 'unevaluatedItems' set to boolean false`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "unevaluatedItems": false
+}
+{{</schema>}}
+
+{{<instance-fail `All object instances fail against the false schema`>}}
+[ "foo", "bar" ]
+{{</instance-fail>}}
+
+{{<instance-pass `'unevaluatedItems' does not have any effect on instances other than an array`>}}
+[ "John", 46, false ]
+{{</instance-pass>}}
+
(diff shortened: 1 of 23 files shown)


content/2020-12/unevaluated/unevaluatedItems.markdown
```diff
@@ -30,5 +30,349 @@ related:
...
+`unevaluatedItems` considers annotations from `prefixItems`, `items`, and `contains`, both as adjacent keywords and in subschemas of adjacent keywords. Additionally, it is also affected by other `unevaluatedItems` in nested schemas (if present).
+
+- The keywords `prefixItems`, `items`, `contains` and `unevaluatedItems` produce annotations for the indexes they successfully validate against.
+- If any of these keywords generate an annotation for a particular index, that index is considered as evaluated.
+- By definition, the `unevaluatedItems` subschema is always applied after  `prefixItems`, `items`, and `contains` subschemas.
+- As its name implies, `unevaluatedItems` applies to any array index that has not been previously evaluated.
+
+## Explanation
+
+If no relevant annotations are present, the `unevaluatedItems` subschema must be applied to all locations in the array. If a boolean true value is present from any of the relevant annotations, `unevaluatedItems` must be ignored. Otherwise, the subschema must be applied to any index greater than the largest annotation value for `prefixItems`, which does not appear in any annotation value for `contains`.
```
jviotti, line 48:
```suggestion
If no relevant annotations are present, the `unevaluatedItems` subschema must be applied to all locations in the array. If a boolean true value is present from any of the relevant annotations, `unevaluatedItems` is ignored. Otherwise, the subschema must be applied to any index greater than the largest annotation value for `prefixItems`, which does not appear in any annotation value for `contains`.
```

Otherwise sounds like the schema writer has to ignore it?

content/2020-12/unevaluated/unevaluatedItems.markdown
```diff
@@ -30,5 +30,349 @@ related:
...
+
+{{<schema `Schema with 'unevaluatedItems' set to boolean true`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "unevaluatedItems ": true
+}
+{{</schema>}}
+
+{{<instance-pass `All object instances pass against the true schema`>}}
+[ "foo": "bar" ]
```
jviotti, line 63: This is not valid JSON. Maybe you missed the curly braces?

content/2020-12/unevaluated/unevaluatedItems.markdown
```diff
@@ -30,5 +30,349 @@ related:
...
+{{</instance-pass>}}
+
+{{<instance-pass `'unevaluatedItems' does not have any effect on instances other than an array`>}}
+"John Doe"
+{{</instance-pass>}}
+
+{{<instance-annotation>}}
```
jviotti, line 70: I think you need to show annotations for the above schema too.

content/2020-12/unevaluated/unevaluatedItems.markdown
jviotti: Ah, I think you just mean `,` instead of `:`

content/2020-12/unevaluated/unevaluatedItems.markdown
```diff
@@ -30,5 +30,349 @@ related:
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "unevaluatedItems": false
+}
+{{</schema>}}
+
+{{<instance-fail `All object instances fail against the false schema`>}}
+[ "foo", "bar" ]
+{{</instance-fail>}}
+
+{{<instance-pass `'unevaluatedItems' does not have any effect on instances other than an array`>}}
```
jviotti, line 95: The fact that the instance is actually an array contradicts this claim?

content/2020-12/unevaluated/unevaluatedItems.markdown
```diff
@@ -30,5 +30,349 @@ related:
...
+{{</instance-pass>}}
+
+{{<instance-annotation>}}
+[
+  // ...
+  {
+    "valid": true,
+    "keywordLocation": "/unevaluatedItems",
+    "instanceLocation": "",
+    "annotation": "true"
```
jviotti, line 77: This annotation is wrong, as the above instance is not an array. It should not have been produced

content/2020-12/unevaluated/unevaluatedItems.markdown
```diff
@@ -30,5 +30,349 @@ related:
...
+    "valid": true,
+    "keywordLocation": "/contains",
+    "instanceLocation": "",
+    "annotation": [ 1, 2 ]
+  },
+  // ...
+]
+{{</instance-annotation>}}
+
+* For the first instance, the annotation result of `prefixItems` is 0, and the annotation result of `contains` is [ 1 ]. However, the item at 2nd index (i.e., `false`) remains unevaluated, so the `unevaluatedItems` subschema applies to it. This subschema fails (as any instance against a false schema is always invalid), leading to the failure of the entire schema.
```
jviotti, line 135:
```suggestion
* For the first instance, the annotation result of `prefixItems` is 0, and the annotation result of `contains` is [ 1 ]. However, the item at 2nd index (i.e., `false`) remains unevaluated, so the `unevaluatedItems` subschema applies to it. This subschema fails (as any instance against a false schema is always invalid), leading to validation failure.
```

content/2020-12/unevaluated/unevaluatedItems.markdown
```diff
@@ -30,5 +30,349 @@ related:
...
+    "valid": true,
+    "keywordLocation": "/unevaluatedItems",
+    "instanceLocation": "",
+    "annotation": true
+  },
+  // ...
+]
+{{</instance-annotation>}}
+
+{{<instance-fail `An array instance with unevaluated items that do not conform to the 'unevaluatedItems' subschema is valid`>}}
```
jviotti, line 200:
```suggestion
{{<instance-fail `An array instance with unevaluated items that do not conform to the 'unevaluatedItems' subschema is invalid`>}}
```

content/2020-12/unevaluated/unevaluatedItems.markdown
```diff
@@ -30,5 +30,349 @@ related:
...
+[ "foo", 101, [ false ] ]
+{{</instance-fail>}}
+
+* For the first instance, there are no unevaluated items.
+
+* For the second instance, the item at 2nd index (i.e., `false`) remains unevaluated, and the `unevaluatedItems` subschema applies to it. This item conforms to this subschema, and hence the instance is valid. The annotations produced by applicators are: `prefixItems` → 0, `contains` → [ 1 ], and `unevaluatedItems` → true.
```
jviotti, line 206: The last sentence seems to be duplicating the annotations already shown in the example. Maybe omit them here?

content/2020-12/unevaluated/unevaluatedItems.markdown
```diff
@@ -30,5 +30,349 @@ related:
...
+    "valid": true,
+    "keywordLocation": "/unevaluatedItems",
+    "instanceLocation": "",
+    "annotation": true
+  },
+  // ...
+]
+{{</instance-annotation>}}
+
+{{<instance-fail `An array instance with unevaluated items that do not conform to the 'unevaluatedItems' subschema is valid`>}}
```
jviotti, line 253:
```suggestion
{{<instance-fail `An array instance with unevaluated items that do not conform to the 'unevaluatedItems' subschema is invalid`>}}
```

jviotti:
I left some comments but looking good!

content/2020-12/unevaluated/unevaluatedItems.markdown
AgniveshChaubey: Actually, the annotations for both instances are same; that's why I skipped one. Okay, let me add for the above one as well.

content/2020-12/unevaluated/unevaluatedItems.markdown
AgniveshChaubey: Ah, yeah. It supposed to be something other than an array.

content/2020-12/unevaluated/unevaluatedItems.markdown
AgniveshChaubey: Actually, this annotation is for the above one only. It just needs to placed above.

content/2020-12/unevaluated/unevaluatedItems.markdown
AgniveshChaubey: Actually, this is for the above instance. It was unknowingly misplaced!

content/2020-12/unevaluated/unevaluatedItems.markdown
AgniveshChaubey: Okay!

AgniveshChaubey:
Updated!
Final submission is going on that's why responded a bit late. Sorry for that.

jviotti (approved):

jviotti:
No worries and good luck!

PR #169, opened 2024-05-07
fix bowtie website link

adwait-godbole:
Thanks @jviotti and @AgniveshChaubey for this awesome resource!


jviotti (approved):
Thank you very much for the fix!

jviotti:
It should be live in a couple of minutes

Issue #52, opened 2024-02-05
Document `$vocabulary`


jviotti:

jviotti:
This one is interesting. We could:

- Showcase how this looks like for the default meta-schema
- Explain the idea of required/optional vocabularies
- Explain that the core vocabulary MUST always be set, and as required
- Explain that `$vocabularies` is not transitive. A meta-schema inheriting from another meta-schema doesn't inherit its vocabularies
- Recommend very much for meta-schemas to declare this keyword

AgniveshChaubey:
This task was addressed in #230.

PR #230, opened 2024-05-17
Keyword `$vocabulary`

AgniveshChaubey:
This PR addresses #52.

I am not very confident with the last three keywords, so will require a bit more guidance on these keywords.


content/2020-12/core/vocabulary.markdown
```diff
@@ -18,3 +18,83 @@ related:
   - vocabulary: core
     keyword: $schema
 ---
+
+Vocabularies are sets of keywords and their meanings used to validate JSON documents. This keyword is crucial for meta-schemas, which are schemas that define other schemas.
```
jviotti, line 22: I think the first sentence is a bit confusing. A vocabulary is an atomic collection of interrelated keywords that can be used in a JSON Schema. These keywords are not only for validation. i.e. you have previously documented annotation keywords like `title`, that come from the "Meta Data" vocabulary

content/2020-12/core/vocabulary.markdown
```diff
@@ -18,3 +18,83 @@ related:
...
+
+The `$vocabulary` keyword is used in meta-schemas to identify the vocabularies available for use in schemas described by that meta-schema. It is also used to indicate whether each vocabulary is required or optional, in the sense that an implementation must understand the required vocabularies in order to successfully process the schema.
+
+* **Required and optional vocabularies:** If a vocabulary is required and an implementation does not recognize it, it must refuse to process any schemas that declare this meta-schema. If a vocabulary is optional, implementations that do not recognize it should proceed with processing such schemas.
+
+* **Mandatory:** The core vocabulary MUST always be included and set as required.
```
jviotti, line 28:
```suggestion
* **Mandatory:** The Core vocabulary MUST always be included and set as required.
```

content/2020-12/core/vocabulary.markdown
```diff
@@ -18,3 +18,83 @@ related:
...
+
+* **Non-inheritability:** Vocabularies defined in one meta-schema do not automatically apply to another meta-schema that references it. Each meta-schema must declare its vocabularies independently.
+
+* **Recommendation:** Meta-schemas should always declare this keyword to clearly specify the vocabularies in use and avoid ambiguities.
+
+## Examples
+
+{{<schema `'$vocabulary' for default meta-schema`>}}
```
jviotti, line 36:
```suggestion
{{<schema `'$vocabulary' for default official 2020-12 meta-schema`>}}
```

content/2020-12/core/vocabulary.markdown
```diff
@@ -18,3 +18,83 @@ related:
...
+  ],
+  // ...
+}
+{{</schema>}}
+
+{{<schema `Custom vocabulary with a custom keyword`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/meta/example-vocab",
+  "$dynamicAnchor": "meta",
```
jviotti, line 66: I think you should explain `$dynamicAnchor: meta` in this page. You can explain it is set by convention to `meta` on the official meta-schemas, and it's a mechanism to enable meta-schema extensibility. By declaring `$dynamicAnchor: meta` here, you are essentially telling JSON Schema to validate EVERY subschema of the instance schema against the meta-schema, and not only the top level of it

content/2020-12/core/vocabulary.markdown
```diff
@@ -18,3 +18,83 @@ related:
...
+  "type": [ "object", "boolean" ],
+  "properties": {
+    "minDate": {
+      "type": "string",
+      "pattern": "\d\d\d\d-\d\d-\d\d",
```
jviotti, line 71: This pattern is invalid JSON. You have to escape the backslashes

content/2020-12/core/vocabulary.markdown
```diff
@@ -18,3 +18,83 @@ related:
...
+    { "$ref": "meta/validation" },
+    { "$ref": "meta/meta-data" },
+    { "$ref": "meta/format-annotation" },
+    { "$ref": "meta/content" }
+  ],
+  // ...
+}
+{{</schema>}}
+
+{{<schema `Custom vocabulary with a custom keyword`>}}
```
jviotti, line 62: This is not a custom vocabulary, but a vocabulary meta-schema

content/2020-12/core/vocabulary.markdown
```diff
@@ -18,3 +18,83 @@ related:
...
+    "minDate": {
+      "type": "string",
+      "pattern": "\d\d\d\d-\d\d-\d\d",
+      "format": "date"
+    }
+  }
+}
+{{</schema>}}
+
+{{<schema `Meta-schema that includes the above vocabulary`>}}
```
jviotti, line 78: Maybe add a variant of this example too importing the example vocabulary as optional? Just to exemplify that

jviotti:
I left a few comments but definitely looks in the right direction. I think you mostly get it already

content/2020-12/core/vocabulary.markdown
AgniveshChaubey: Okay, lets omit that

content/2020-12/core/vocabulary.markdown
AgniveshChaubey: can you provide an example of a custom vocabulary?

content/2020-12/core/vocabulary.markdown
AgniveshChaubey: Should I add it under <learning-more> block or just below that schema block as a note?

AgniveshChaubey: Should I add it under `leaning-more` block or just below that schema block as a note?

jviotti: Yeah, that's a good point. Let's use the new `learning-more` block :)

content/2020-12/core/vocabulary.markdown
```diff
@@ -18,3 +18,99 @@ related:
...
+    "https://example.com/vocab/example-vocab": "true"
+  },
+  "allOf": [
+    { "$ref": "https://json-schema.org/draft/2020-12/meta/core" },
+    { "$ref": "https://example.com/meta/example-vocab" }
+  ]
+}
+{{</schema>}}
+
+{{<schema `Meta-schema with the above vocabularyas optional`>}}
```
jviotti, line 94:
```suggestion
{{<schema `Meta-schema with the above vocabulary as optional`>}}
```

PR #233, opened 2024-05-18
Alphabetical ordering in the sidebar

Era-cell:
Addressing this issue: https://github.com/Intelligence-AI/learnjsonschema.com/issues/36
I need assisstance in positioning and naming the toggle bar


layouts/partials/sidebar.html
```diff
@@ -1,37 +1,65 @@
...
+            {{ $keywordURL := (replace .Params.keyword "$" "") | printf "%s/%s" $vocabularyURL }}
+            <li>
+              {{ if eq $.Permalink .Permalink }}
+              <a
+                class="text-muted fw-bold text-decoration-none font-monospace active"
+                href="{{ ref $ $keywordURL }}"
+                aria-current="true">{{ .Params.keyword }}</a>
+              {{ else }}
+              <a
+                class="text-black-50 text-decoration-none font-monospace"
```
jviotti, line 56: I think this is regressing on a similar vein as https://github.com/Intelligence-AI/learnjsonschema.com/pull/237. Can you double-check it renders well on both light and dark theme?

layouts/partials/sidebar.html
```diff
@@ -1,37 +1,65 @@
...
+              <ul class="list-unstyled ms-3">
+                {{ range sort $.Site.RegularPages "Params.index" "asc" }}
+                  {{ if eq .Section $dialect }}
+                    {{ if eq .Parent.Title $vocabulary }}
+                      {{ $keywordURL := (replace .Params.keyword "$" "") | printf "%s/%s" $vocabularyURL }}
+                      <li>
+                        {{ if eq $.Permalink .Permalink }}
+                          <a class="text-muted fw-bold text-decoration-none font-monospace active" href="{{ ref $ $keywordURL }}" aria-current="true">{{ .Params.keyword }}</a>
+                        {{ else }}
+                          <a class="text-black-50 text-decoration-none font-monospace" href="{{ ref $ $keywordURL }}">{{ .Params.keyword }}</a>
```
jviotti, line 31: This here too. Maybe will look bad on dark theme

assets/main.scss
```diff
@@ -141,3 +141,16 @@ article table {
...
+details:not([open]) summary i {
+  @extend .d-inline-block;
+  transform: rotate(-90deg); 
+}
+
+.keyword-list {
+  @extend .d-block
+}
+
+.form-check:has(#flexSwitchCheckDefault:checked) ~ .keyword-list{
```
jviotti, line 154: Not sure the `.keyword-list` class is contributing much here. Could we do something like this instead?

```suggestion
.form-check:has(#flexSwitchCheckDefault:checked) ~ ul {
```

jviotti: Also mind the space before `{`!

layouts/partials/sidebar.html
```diff
@@ -1,37 +1,65 @@
...
+  
+  <div class="px-4 py-2">
+    {{ $selected_dialect := index (split .RelPermalink "/") 1 }}
     {{ range .Site.Sections }}
       {{ $dialect := .Title }}
       {{ $dialectURL := $dialect | printf "/%s" }}
-      <ul class="nav flex-column text-nowrap mb-4">
+      <div class="form-check form-switch">
+        <label class="form-check-label" for="flexSwitchCheckDefault">Toggle Display</label>
+        <input class="form-check-input" type="checkbox" role="switch" id="flexSwitchCheckDefault">
```
jviotti, line 10: This `id` feels weird. Maybe it was a placeholder example from Bootstrap? Maybe call it something like `sidebarOrdering`?

layouts/partials/sidebar.html
```diff
@@ -1,37 +1,65 @@
...
-  <div class="px-4 py-2 top-0 start-0 bottom-0">
+  
+  <div class="px-4 py-2">
+    {{ $selected_dialect := index (split .RelPermalink "/") 1 }}
     {{ range .Site.Sections }}
       {{ $dialect := .Title }}
       {{ $dialectURL := $dialect | printf "/%s" }}
-      <ul class="nav flex-column text-nowrap mb-4">
+      <div class="form-check form-switch">
+        <label class="form-check-label" for="flexSwitchCheckDefault">Toggle Display</label>
```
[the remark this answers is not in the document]
jviotti: Just for brainstorming purposes, maybe another way of doings it would be with a tab (https://getbootstrap.com/docs/5.3/components/navs-tabs/#tabs) or pill (https://getbootstrap.com/docs/5.3/components/navs-tabs/#pills) menu, where one is "By Vocabulary" and the other is "Alphabetically"?

layouts/partials/sidebar.html
jviotti: Also, it'd be good to make the browser remember this selection. Can write a very short snippet of JavaScript in `main.js` that caches the value set here to `localStorage` (and restores it on load if it was set)?

assets/main.scss
Era-cell: I need this class only because I need to apply functionality only for the first ul. Is it ok to have it?

layouts/partials/sidebar.html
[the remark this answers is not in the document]
Era-cell: I think we can go with pills, if I am allowed to write 10-15 lines of Javascript or CSS for it?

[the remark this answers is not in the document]
jviotti: Yeah, I think that's fine. You can keep the selection in `localStorage`

Era-cell:
@jviotti  I have tried to add pills instead of toggle bar. Looks very naive, but I could do this much.. Maybe we can optimize it idk.
Have a first look on it, ...

jviotti:
Yeah, I think the toggle was way better

Era-cell:
> Yeah, I think the toggle was way better

I see that the behavior of flickering is happening due to setting up the sidebar based on the localStorage data, which all this is happening after the page loads. How can this be solved..

jviotti:
What flickering are you seeing? Can you show me? Is that on the current changes?

Era-cell:
> What flickering are you seeing? Can you show me? Is that on the current changes?

Yeah! thats when I and retrieving the state from localStorage and then trying to use 

Uploading Recording 2024-05-23 132113.mp4…

jviotti:
I don't think you uploaded any video. In any case, if we are going back to the toggle, then no point in debugging the pills approach? (the pills don't even work for me at the moment)

Era-cell:
> I don't think you uploaded any video. In any case, if we are going back to the toggle, then no point in debugging the pills approach? (the pills don't even work for me at the moment)

I think the video didn't get uploaded. Its weird that it isn't working, anyway i'll upload the video:

https://github.com/Intelligence-AI/learnjsonschema.com/assets/77973415/e86ea531-893e-4aeb-97b9-ee6008f44aef

The flickering remains the same for toggle too. So, if we resolve it then having pills will look good..

jviotti:
Ah, I think it's because the JavaScript file runs in deferred mode, once the page is already loaded. In this case it might make sense to add a blocking inline `<script>` near the toggle

Era-cell:
> Ah, I think it's because the JavaScript file runs in deferred mode, once the page is already loaded. In this case it might make sense to add a blocking inline `<script>` near the toggle

@jviotti it works well I guess you can check now, and I assume it works with light mode too, I don't know hoe do I check with that

assets/main.scss
```diff
@@ -141,3 +141,16 @@ article table {
     @extend .bg-info-subtle;
   }
 }
+
+details:not([open]) summary i {
```
jviotti, line 145: Wasn't this from the other PR?

assets/main.scss
jviotti: Not sure I get it 100%, but could you use `:first-child` or something like that?

layouts/partials/sidebar.html
```diff
@@ -1,37 +1,106 @@
...
-        {{ end }}
-      </ul>
+            {{ end }}
+          </ul>
+        </div>
+      </div>
     {{ end }}
   </div>
 </aside>
+<script>
```
jviotti, line 70: I meant you can add an inline script with maybe a one liner to restore the value from local storage, but the rest of it, as much as possible, should be in `main.js`

layouts/partials/sidebar.html
```diff
@@ -1,37 +1,106 @@
...
-                  {{ if eq .Parent.Title $vocabulary }}
-                    {{ $keywordURL := (replace .Params.keyword "$" "") | printf "%s/%s" $vocabularyURL }}
-                    <li>
-                      {{ if eq $.Permalink .Permalink }}
-                        <a class="text-muted fw-bold text-decoration-none font-monospace active"
-                          href="{{ ref $ $keywordURL }}" aria-current="true">{{ .Params.keyword }}</a>
-                      {{ else }}
-                        <a class="text-secondary text-decoration-none font-monospace"
-                          href="{{ ref $ $keywordURL }}">{{ .Params.keyword }}</a>
+      <ul class="nav nav-pills" id="sidebarNavPills">
```
jviotti, line 7: Is this id really needed? Also, usually HTML identifies are in snake-case, instead of camelCase

layouts/partials/sidebar.html
```diff
@@ -1,37 +1,106 @@
...
+        <li class="nav-item me-2">
+          <a class="nav-link px-2 active" id="vocabulary-tab" data-bs-toggle="pill" href="#">Vocabulary</a>
```
jviotti, line 9: The fact that these links are indeed anchors to `#` is resulting in the empty fragment getting appended to the URL, but I guess all of this might be fixed if we go back to the toggle approach?

layouts/partials/sidebar.html
```diff
@@ -1,37 +1,106 @@
...
+      alphabeticTab.classList.remove('active');
+      vocabularyTab.classList.add('active');
+      alphabeticContent.classList.remove('show', 'active');
+      alphabeticContent.classList.add('d-none');
+      vocabularyContent.classList.add('show', 'active');
+      vocabularyContent.classList.remove('d-none');
+    }
+  }
+
+  const alphabeticOrder = sessionStorage.getItem('alphabeticOrder') === 'true';
```
jviotti, line 94: You probably want local storage instead of session storage!

layouts/partials/sidebar.html
```diff
@@ -1,37 +1,106 @@
...
+  setActiveTab(alphabeticOrder ? 'alphabetic' : 'vocabulary');
+
+  vocabularyTab.addEventListener('click', function () {
+    sessionStorage.setItem('alphabeticOrder', 'false');
```
jviotti, line 98: The storage entry feels weird if it's just about whether alphabetic order is set or not. It would also break down if we added one more order type. Also, it doesn't say what the ordering is for. If we add alphabetic ordering to another part of the site, it would also break down.

I recommend setting to something like `sidebarGrouping` = `alphabetic` or `vocabulary`

layouts/partials/sidebar.html
```diff
@@ -1,37 +1,106 @@
...
     {{ end }}
   </div>
 </aside>
+<script>
+  const vocabularyTab = document.getElementById('vocabulary-tab');
+  const alphabeticTab = document.getElementById('alphabetic-tab');
+  const vocabularyContent = document.getElementById('vocabulary');
+  const alphabeticContent = document.getElementById('alphabetic');
+
+  function setActiveTab(tab) {
```
jviotti, line 76: This function is on the global scope, and it's not very clear on what tabs it's even referring to. Maybe call it `setSidebarGrouping` or something more meaningful?

layouts/partials/sidebar.html
```diff
@@ -1,37 +1,106 @@
...
+    if (tab === 'alphabetic') {
```
jviotti, line 77: I think you can simplify this logic a lot, but let's do the rest first.

jviotti:
I dropped a few comments but it's shaping up! Also, there are git conflicts on the PR. Can you fix those?

layouts/partials/sidebar.html
jviotti: I was not very convinced about how the pills work, but there is a proposal: we can render the same nav you have as tabs. We can make it centred, reduce the font size a bit and tweak spacing to make it looks like this:

Screenshot 2024-05-29 at 8 49 30 PM

I think it looks quite good. What do you think? These where the changes I did:

```diff
diff --git a/assets/main.scss b/assets/main.scss
index d1969fd..e149e3e 100644
--- a/assets/main.scss
+++ b/assets/main.scss
@@ -6,6 +6,7 @@ $danger: rgba(255, 213, 213, 1);
 $navbar-brand-font-size: 1.1rem;
 $navbar-padding-y: 1px;
 $alert-border-radius: 0;
+$nav-link-padding-y: 0.5rem;
 
 // Follow browser's preferred theme
 $color-mode-type: media-query;
@@ -74,7 +75,11 @@ $bootstrap-icons-font-dir: "./";
 // Other
 
 #sidebar {
-  min-width: 260px;
+  min-width: 280px;
+}
+
+#sidebarNavPills {
+  font-size: $font-size-sm * 0.8;
 }
 
 // Tables with vertical headers collapse the headers by default
@@ -144,7 +149,7 @@ article table {
 
 details:not([open]) summary i {
   @extend .d-inline-block;
-  transform: rotate(-90deg); 
+  transform: rotate(-90deg);
 }
 
 .keyword-list {
diff --git a/layouts/partials/sidebar.html b/layouts/partials/sidebar.html
index 7dfa096..fc6e5e7 100644
--- a/layouts/partials/sidebar.html
+++ b/layouts/partials/sidebar.html
@@ -4,12 +4,16 @@
     {{ range .Site.Sections }}
       {{ $dialect := .Title }}
       {{ $dialectURL := $dialect | printf "/%s" }}
-      <ul class="nav nav-pills" id="sidebarNavPills">
-        <li class="nav-item me-2">
-          <a class="nav-link px-2 active" id="vocabulary-tab" data-bs-toggle="pill" href="#">Vocabulary</a>
+      <ul class="nav nav-tabs nav-fill mt-1" id="sidebarNavPills">
+        <li class="nav-item">
+          <a class="nav-link active" id="vocabulary-tab" data-bs-toggle="pill" href="#">
+            By Vocabulary
+          </a>
         </li>
-        <li class="nav-item me-2">
-          <a class="nav-link px-2" id="alphabetic-tab" data-bs-toggle="pill" href="#">Alphabetic</a>
+        <li class="nav-item">
+          <a class="nav-link" id="alphabetic-tab" data-bs-toggle="pill" href="#">
+            Alphabetically
+          </a>
         </li>
       </ul>
       <div class="tab-content">
```

Era-cell: Yeah! this looks better than having complete pills. I thought of doing something like this, but didn't know t would be that simple and can be done using just bootstrap classes. From where do I learn and refer to it..?

jviotti: I really recommend reading the Bootstrap documentation front to back: https://getbootstrap.com/docs/5.3/getting-started/introduction/, going through every page in the left sidebar. It explains everything about Bootstrap and how to use it correctly (as most people use it incorrectly) compared to any other resources

Era-cell:
@jviotti  can you review this PR, functionality is good

jviotti:
@Era-cell Looks good, but after the addition of 2019-09, the sidebar seems to be showing the nav twice, once at the top, and once at the bottom

Screenshot 2024-06-19 at 9 45 22 AM



Screenshot 2024-06-19 at 9 45 30 AM

Issue #54, opened 2024-02-05
Document `$dynamicAnchor`


jviotti:

jviotti:
OK, this is a tricky one. I was planning to write a blog post about it as a follow-up to my scoping blog post. Maybe let's keep this keyword to the end, and we can more easily write docs about it once by post is up?

AgniveshChaubey:
This task was addressed in #235.

PR #235, opened 2024-05-20
Keyword `$dynamicAnchor`

AgniveshChaubey:
This PR addresses #54.


content/2020-12/core/dynamicAnchor.markdown
```diff
@@ -21,3 +21,50 @@ related:
   - vocabulary: core
     keyword: $anchor
 ---
+
+The `$dynamicAnchor` keyword allows the creation of plain name fragments that are not tied to a particular structural location within a schema. This is particularly useful for making subschemas reusable and relocatable without needing to update JSON Pointer references. Unlike `$anchor`, `$dynamicAnchor` indicates an extension point when used with the `$dynamicRef` keyword, facilitating the extension of recursive schemas without imposing specific semantics on that extension. Without `$dynamicRef`, `$dynamicAnchor` behaves the same as `$anchor`.
+
+## Examples
+
+{{<schema `Schema with '$dynamicRef' and '$dynamicAnchor' keywords in the same schema resource`>}}
```
jviotti, line 29: I think you should use this schema as an example of a schema with `$ref` and `$dynamicAnchor`, and have a new example that actually makes use of the dynamic scope. Maybe adapt one of the JSON Schema Test Suite examples that use `$dynamicRef` and `$dynamicAnchor`?

content/2020-12/core/dynamicAnchor.markdown
```diff
@@ -21,3 +21,50 @@ related:
...
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/1",
+  "required": [ "name", "age", "address" ],
+  "properties": {
+    "name": { "$dynamicRef": "#name" },
+    "age": { "$dynamicRef": "#age" },
+    "address": { "$ref": "#address" },
+    "$defs": {
```
jviotti, line 38: You are putting `$defs` inside `properties`

content/2020-12/core/dynamicAnchor.markdown
```diff
@@ -21,3 +21,50 @@ related:
...
 ---
+
+The `$dynamicAnchor` keyword allows the creation of plain name fragments that are not tied to a particular structural location within a schema. This is particularly useful for making subschemas reusable and relocatable without needing to update JSON Pointer references. Unlike `$anchor`, `$dynamicAnchor` indicates an extension point when used with the `$dynamicRef` keyword, facilitating the extension of recursive schemas without imposing specific semantics on that extension. Without `$dynamicRef`, `$dynamicAnchor` behaves the same as `$anchor`.
+
+## Examples
+
+{{<schema `Schema with '$dynamicRef' and '$dynamicAnchor' keywords in the same schema resource`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/1",
```
jviotti, line 32: No reason to put an id here, as you are not using it

content/2020-12/core/dynamicAnchor.markdown
```diff
@@ -21,3 +21,50 @@ related:
...
+  "age": 35,
+  "address": "1234 Elm Street, Springfield, IL 62701, USA"
+}
+{{</instance-pass>}}
+
+{{<instance-fail `Required properties must be present`>}}
+{ "name": "Doe", "age": 61 }
+{{</instance-fail>}}
+
+* _A `$dynamicRef` referencing a `$dynamicAnchor` within the same schema resource functions similarly to a standard `$ref` referencing an `$anchor`. Similarly, a `$dynamicRef` referencing an `$anchor` within the same schema resource behaves like a typical `$ref` referencing an `$anchor`. Likewise, a `$ref` targeting a `$dynamicAnchor` within the same schema resource behaves like a regular `$ref` targeting an `$anchor`._
```
jviotti, line 70: I think you need an example that actually makes use of the dynamic scope.

AgniveshChaubey:
Updated

AgniveshChaubey:
Maybe it's worth adding the bookending explanation here as well?

content/2020-12/core/dynamicAnchor.markdown
```diff
@@ -21,3 +21,44 @@ related:
...
+    },
+    "list": {
+      "$id": "list",
+      "type": "array",
+      "items": {
+        "$dynamicRef": "#items"
+      },
+      "$defs": {
+        "items": {
+          "$comment": "This is only needed to satisfy the bookending requirement",
```
jviotti, line 47: Maybe instead of mentioning bookending here, add a `type: number` or something like that to contrast it with the other destination?

jviotti (approved):
I left a small suggestion but looks good. I think we can avoid mentioning bookending here given what I suggested, and then we only cover it on `dynamicRef`

PR #236, opened 2024-05-20
Keyword `$dynamicRef`

AgniveshChaubey:
This PR addresses #55.


content/2020-12/core/dynamicRef.markdown
```diff
@@ -23,3 +23,183 @@ related:
...
+The `$dynamicRef` keyword is a dynamic applicator that allows for runtime resolution of schema references. Unlike the static `$ref`, which resolves the referenced schema at schema load time, `$dynamicRef` defers full resolution until the instance is evaluated. This keyword is particularly useful for handling recursive schemas where the schema references itself or where the structure of the schema may change at runtime.
+
+{{<learning-more>}} URIs play a central role in JSON Schema. Going through the
+URI [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) specification is a
+must for gaining a deeper understanding of references, identifiers, and
+anchors. More specifically, we recommend carefully studying [URI
+resolution](https://datatracker.ietf.org/doc/html/rfc3986#section-5), URLs vs
+URNs, and the difference between a URI and a URI Reference.
+
+Additionally, a JSON Schema reference URI may contain a JSON Pointer. For this
```
jviotti, line 36: This paragraph is not applicable to anchors

content/2020-12/core/dynamicRef.markdown
```diff
@@ -23,3 +23,183 @@ related:
...
+{{<schema `Schema with '$dynamicRef' and '$dynamicAnchor' keywords in the same schema resource`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/1",
+  "required": [ "name", "age", "address" ],
+  "properties": {
+    "name": { "$dynamicRef": "#name" },
+    "age": { "$dynamicRef": "#age" },
+    "address": { "$ref": "#address" },
+    "$defs": {
```
jviotti, line 54: Also as I said on the other PR, this `$defs` is invalid and the `$id` is unused

content/2020-12/core/dynamicRef.markdown
```diff
@@ -23,3 +23,183 @@ related:
...
+{{<instance-fail `Required properties must be present`>}}
+{ "name": "Doe", "age": 61 }
+{{</instance-fail>}}
+
+* _A `$dynamicRef` referencing a `$dynamicAnchor` within the same schema resource functions similarly to a standard `$ref` referencing an `$anchor`. Similarly, a `$dynamicRef` referencing an `$anchor` within the same schema resource behaves like a typical `$ref` referencing an `$anchor`. Likewise, a `$ref` targeting a `$dynamicAnchor` within the same schema resource behaves like a regular `$ref` targeting an `$anchor`._
+
+{{<schema `Schema with multiple '$dynamicAnchor' set to same value`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/2",
```
jviotti, line 91: `$id` is unused here too. Let's not put `$id` if examples don't rely on it, like for other keywords you documented

jviotti, line 91: Let's use a more semantic identifier here rather than just "2". Maybe "top-level" or "root"?

content/2020-12/core/dynamicRef.markdown
```diff
@@ -23,3 +23,183 @@ related:
...
+  "$ref": "list",
+  "$defs": {
+    "foo": {
```
jviotti, line 94: I'm not sure I follow this example. There is a `foo` that is completely unused and a dynamic ref to an anchor that would only resolve to one place (as this anchor it outside the other schema resource)

content/2020-12/core/dynamicRef.markdown
```diff
@@ -23,3 +23,183 @@ related:
...
+{{</instance-fail>}}
+
+{{<schema `Schema with '$anchor' and '$dynamicAnchor' set to same value`>}}
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/3",
+  "$ref": "list",
+  "$defs": {
+    "foo": {
+      "$anchor": "items",
```
jviotti, line 127: This example doesn't really show anything, as this anchor applies to the identifier of the root schema resource and wouldn't be visible to the embedded schema resource anyway

content/2020-12/core/dynamicRef.markdown
```diff
@@ -23,3 +23,183 @@ related:
...
+  "foo": "pass",
+  "bar": {
+    "baz": { "foo": "fail" }
+  }
+}
+{{</instance-fail>}}
+
+* _A `$dynamicRef` that initially resolves to a schema with a matching `$dynamicAnchor` resolves to the first `$dynamicAnchor` in the dynamic scope._
+
+{{<learning-more>}}
```
jviotti, line 201: Let's put this alert when describing the keyword. Otherwise nobody will find it down here

jviotti (changes requested):
There are various important parts of dynamic references that are not being explained here. I'd expect:

- A mention of the "bookending" requirement and what that means
- A dynamic reference that contains more than just a fragment (i.e. `$dynamicRef: "foo#bar"`) and how that is resolved (non-fragment is always statically resolved and the fragment is maybe dynamically resolved)
- At least a mention to scopes in the description. i.e. that `$dynamicRef` will attempt to resolve the given fragment based on the dynamic scope at that given point in time

I think this is the best example from the test case that you can adapt and discuss: https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft2020-12/dynamicRef.json#L477-L513. It shows many of the above things already in a single example

content/2020-12/core/dynamicRef.markdown
```diff
@@ -23,3 +23,183 @@ related:
...
+[ 11, 22 ]
+{{</instance-pass>}}
+
+{{<instance-fail `A non-integer array is invalid`>}}
+[ "foo", "bar" ]
+{{</instance-fail>}}
+
+* _A `$dynamicRef` resolves to the first `$dynamicAnchor` still in scope that is encountered when the schema is evaluated. Refer to the above two examples for clarification._
+
+{{<schema `Schema with '$dynamicAnchor' and '$dynamicRef' keyword`>}}
```
jviotti, line 154: I'd replace this example with an adaptation of the `if`/`then`/`else` example I suggested in the other comment, and put it as the first example in the page.

jviotti:
I think you could mention the bookending requirement as a "common pitfall" alert. Most people miss that

content/2020-12/core/dynamicRef.markdown
```diff
@@ -23,3 +23,183 @@ related:
   - vocabulary: core
     keyword: $defs
 ---
+
+The `$dynamicRef` keyword is a dynamic applicator that allows for runtime resolution of schema references. Unlike the static `$ref`, which resolves the referenced schema at schema load time, `$dynamicRef` defers full resolution until the instance is evaluated. This keyword is particularly useful for handling recursive schemas where the schema references itself or where the structure of the schema may change at runtime.
```
jviotti, line 27: I think it'd also be fun to add one line to this description stating that official meta-schemas use this mechanism themselves for defining vocabularies!

content/2020-12/core/dynamicRef.markdown
AgniveshChaubey: This is what I wanted to demonstrate with this example, that only `$dynamicAnchor` will be recognized in the embedded schema, not the `$anchor` and the `$dynamicRef` will be resolved to the first `$dynmicAnchor` encountered when the schema is resolved.

AgniveshChaubey:
Updated

1. Let me know if the description for bookending makes sense.

2. Taking referece from your blog post, I have explained that if/else example in detail, so that one can get good enough understanding of fynamic scoping here itself. Let me know if should should be included here or not.

content/2020-12/core/dynamicRef.markdown
```diff
@@ -23,3 +23,168 @@ related:
...
+{{<learning-more>}} URIs play a central role in JSON Schema. Going through the URI [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) specification is a must for gaining a deeper understanding of references, identifiers, and
+anchors. More specifically, we recommend carefully studying [URI resolution](https://datatracker.ietf.org/doc/html/rfc3986#section-5), URLs vs URNs, and the difference between a URI and a URI Reference.
+
+You may also find these blog posts helpful for gaining a deeper understanding of dynamic references.
+* [Understanding JSON Schema Lexical and Dynamic Scopes](https://json-schema.org/blog/posts/understanding-lexical-dynamic-scopes)
+* [Using Dynamic References to Support Generic Types](https://json-schema.org/blog/posts/dynamicref-and-generics)
+{{</learning-more>}}
+
+{{<common-pitfall>}}
+**Bookending:** The bookending requirement means that when you use a `$dynamicRef`, the JSON Schema processor needs to find a matching `$dynamicAnchor` within the same schema scope to resolve the reference correctly. This ensures that the reference doesn't end up being unresolvable due to scope issues.
```
jviotti, line 38: I think a better way to explain it is that it is expected that the target schema resource of a dynamic reference (in case it's different than the destination) defines that actual anchor.

content/2020-12/core/dynamicRef.markdown
```diff
@@ -23,3 +23,168 @@ related:
   - vocabulary: core
     keyword: $defs
 ---
+
+The `$dynamicRef` keyword is a dynamic applicator that allows for runtime resolution of schema references. Unlike the static `$ref`, which resolves the referenced schema at schema load time, `$dynamicRef` defers full resolution until the instance is evaluated. It attempts to resolve the given fragment based on the dynamic scope at that given point in time. This keyword is particularly useful for handling recursive schemas where the schema references itself or where the structure of the schema may change at runtime. Notably, official meta-schemas use this mechanism themselves for defining vocabularies!
```
jviotti, line 27: One important thing to explain here is that if dynamic ref is not just a fragment, then the URI _except_ the fragment is statically resolved, and then only the fragment is dynamically resolved

content/2020-12/core/dynamicRef.markdown
```diff
@@ -23,3 +23,168 @@ related:
...
+    "foo": {
+      "$dynamicAnchor": "items",
+      "type": "string"
+    },
+    "list": {
+      "$id": "list",
+      "items": { "$dynamicRef": "#items" },
+      "$defs": {
+        "items": {
+          "$comment": "This is only needed to satisfy the bookending requirement",
```
jviotti, line 172: This example is specifically about multiple dynamic anchors, so this comment is a lie: it's not only about the bookending requirement, but an integral part of the example? I say you leave the comment out!

AgniveshChaubey:
Updated!

jviotti (approved):

PR #237, opened 2024-05-21
Fix sidebar text-color in dark mode

AgniveshChaubey:
Before:
image

After:
image


jviotti:
Ahhh I think this got regressed from @Era-cell 's recent PR on the sidebar as a logical conflict. Thanks for fixing it!

PR #238, opened 2024-05-23
Changed in field for all the keywords

Era-cell:
Hi, I have added the changed in attributes, please review and ask me if any doubt in it

### DIFF
--- a/content/2020-12/applicator/items.markdown
+++ b/content/2020-12/applicator/items.markdown
@@ -14,7 +14,6 @@ tests:
 introduced_in: draft1
 changed_in:
   - draft6
-  - 2020-12
 annotation:
    description: A boolean true if it applied to any item of the instance
    kind: [ "boolean" ]
--- a/content/2020-12/applicator/items.markdown
+++ b/content/2020-12/applicator/items.markdown
@@ -12,6 +12,9 @@ default:
 tests:
   - draft2020-12/items.json
 introduced_in: draft1
+changed_in:
+  - draft6
+  - 2020-12
 annotation:
    description: A boolean true if it applied to any item of the instance
    kind: [ "boolean" ]
--- a/content/2020-12/core/anchor.markdown
+++ b/content/2020-12/core/anchor.markdown
@@ -10,8 +10,6 @@ metaschema: "https://json-schema.org/draft/2020-12/meta/core"
 tests:
   - draft2020-12/anchor.json
 introduced_in: 2019-09
-changed_in:
-  - 2020-12
 affects:
   - vocabulary: core
     keyword: $ref
--- a/content/2020-12/core/anchor.markdown
+++ b/content/2020-12/core/anchor.markdown
@@ -10,6 +10,8 @@ metaschema: "https://json-schema.org/draft/2020-12/meta/core"
 tests:
   - draft2020-12/anchor.json
 introduced_in: 2019-09
+changed_in:
+  - 2020-12
 affects:
   - vocabulary: core
     keyword: $ref
--- a/content/2020-12/format-annotation/format.markdown
+++ b/content/2020-12/format-annotation/format.markdown
@@ -13,8 +13,9 @@ introduced_in: draft1
 changed_in:
   - draft3
   - draft4
+  - draft6
+  - draft7
   - 2019-09
-  - 2020-12
 annotation:
    description: The format name set by this keyword
    kind: [ "string" ]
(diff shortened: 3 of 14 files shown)


content/2020-12/applicator/items.markdown
```diff
@@ -12,6 +12,9 @@ default:
 tests:
   - draft2020-12/items.json
 introduced_in: draft1
+changed_in:
+  - draft6
+  - 2020-12
```
jviotti, line 17: You are documenting 2020-12 itself, so you shouldn't be mentioning 2020-12 here!  Only Draft 6 in this case. Please remove 2020-12 from all keywords here!

content/2020-12/applicator/items.markdown
```diff
@@ -12,6 +12,9 @@ default:
 tests:
   - draft2020-12/items.json
 introduced_in: draft1
+changed_in:
+  - draft6
```
jviotti, line 16: I'm not sure this looks good. Going through the specs:

- Syntactically, `items` remained the same (accepting either an array of schemas or a schema) since Draft 0 up to 2019-09. Then 2020-12 make it only a schema
- Semantically, it was the same until 2019-09 started making it produce an annotation

So you should definitely have 2019-09 in this array, plus not sure what changed in Draft 6 (can't find anything myself)

content/2020-12/core/anchor.markdown
```diff
@@ -10,6 +10,8 @@ metaschema: "https://json-schema.org/draft/2020-12/meta/core"
 tests:
   - draft2020-12/anchor.json
 introduced_in: 2019-09
+changed_in:
+  - 2020-12
```
jviotti, line 14: These are entries for 2020-12 already. A keyword cannot be changed by itself

content/2020-12/format-annotation/format.markdown
```diff
@@ -10,6 +10,11 @@ metaschema: "https://json-schema.org/draft/2020-12/meta/format-annotation"
 tests:
   - draft2020-12/format.json
 introduced_in: draft1
+changed_in:
```
jviotti, line 13: But according to https://github.com/json-schema-org/upgrade-downgrade-rules/pull/6, there were also changes in Draft 6 and Draft 7?

content/2020-12/format-assertion/format.markdown
```diff
@@ -30,6 +30,11 @@ tests:
   - draft2020-12/optional/format/uri.json
   - draft2020-12/optional/format/uuid.json
 introduced_in: draft1
+changed_in:
```
jviotti, line 33: Same as above

content/2020-12/format-annotation/format.markdown
Era-cell: I have included only behavioral changes across the drafts, as you suggested the changes in values inside isnt needed to be documented.
Like for `$vocabulary` the values like "http://..../format-assertion" ~ isnt this same to the `format` if the change in value inside should be considered..

content/2020-12/applicator/items.markdown
Era-cell: From 2019 to 2020. `items` can have array of JSON Schemas or JSON Schema was changed to have only JSON Schema..this one.
I got it, you mean because we are having the whole information of `items` and its behavior in of 2020-12 in the learnjosnschema/2020 site. You mean its not necessary to have it say it was changed in 2020-12??

content/2020-12/applicator/items.markdown
Era-cell: One change across set of keywords in draft6 was: Any keyword taking schema can have boolean values, that was the change

jviotti: Ah, fine then

content/2020-12/applicator/items.markdown
jviotti: Correct! We should only add references to 2019-09 or lower!

content/2020-12/format-annotation/format.markdown
jviotti: I think it still counts as changes within the keyword. For example, the same keyword being moved across vocabularies is outside the keyword, but if the valid values to set the keyword to have changed, then that's a change inside the keyword.

Era-cell: Okay, got it.. lets include them then👍

Era-cell:
@jviotti I have updated the changes

jviotti (approved):
Awesome!

PR #241, opened 2024-05-26
Fix ugly side markers in Safari for `<details>` and `<summary>`

jviotti:


jviotti:
For example:

Screenshot 2024-05-26 at 21 38 58

PR #244, opened 2024-05-30
Link the keywords to their respective pages

AgniveshChaubey:
Addresses #188.


jviotti:
Looks great already @AgniveshChaubey ! One small thing: because in the future we might want to link to keywords from older dialects, etc:

- Can we extend the shortcode to take 2 optional arguments: the dialect name and the vocabulary?
- If those are not passed, you default to the current dialect as you are doing already

content/2020-12/applicator/additionalProperties.markdown
```diff
@@ -41,9 +41,9 @@ related:
     keyword: unevaluatedProperties
 ---
 
-The `additionalProperties` keyword is used to control the handling of properties whose names are not listed in the `properties` keyword or match any of the regular expressions in the `patternProperties` keyword. By default any additional properties are allowed.
+The  {{<keyword-link name="additionalProperties" >}} keyword is used to control the handling of properties whose names are not listed in the  {{<keyword-link name="properties" >}} keyword or match any of the regular expressions in the  {{<keyword-link name="patternProperties" >}} keyword. By default any additional properties are allowed.
```
jviotti, line 44: Can we always require a vocabulary name when using the shortcode? As we might be linking to keywords in other vocabularies?

layouts/shortcodes/keyword-link.html
```diff
@@ -0,0 +1,15 @@
...
+{{- range $.Site.RegularPages -}}
+  {{- $dialect := .Section -}}
+  {{- $vocabulary := .Parent.Title -}}
+  {{- $keywordParam := .Params.keyword -}}
+  {{- $url := .Permalink -}}
+  {{- $keywordMap = $keywordMap | merge (dict $keywordParam $url) -}}
+{{- end -}}
+{{- $url := index $keywordMap $keyword -}}
+{{- if $url -}}
+  <a href="{{ $url }}">{{ $keyword }}</a>
```
jviotti, line 12: Hugo has a `ref` shortcode that ensures the destination is a valid page (thus protecting against broken links): https://gohugo.io/content-management/cross-references/. Can you make use of those here?

AgniveshChaubey:
I am a bit busy these days, so I haven't had much time to work on this. Let the PR remain open; I'll finish it up whenever I get some free time.

jviotti:
Yeah, no hurries!

PR #246, opened 2024-06-06
Adding index json for data generation used in fuse search

Era-cell:


jviotti:
Don't you need the site config changes too for the JSON output to be generated?

layouts/_default/index.json
```diff
@@ -0,0 +1,8 @@
+{{- $.Scratch.Add "index" slice -}}
+{{- range .Site.RegularPages -}}
+    {{ $parts := split  .Permalink  "/" }}
+    {{ $dialect := index $parts 3 }}
+    {{ $vocabulary := index $parts 4 }}
+    {{- $.Scratch.Add "index" (dict "dialect" $dialect "vocabulary" $vocabulary "title" .Params.keyword "tags" .Params.kind "contents" .Plain "permalink" .Permalink) -}}
+{{- end -}}
+{{- $.Scratch.Get "index" | jsonify -}}
```
jviotti, line 8: I think this file is missing a line ending at the end, so GitHub is showing this alert circle.

Era-cell:
> Don't you need the site config changes too for the JSON output to be generated?

Correct. Let me add it

layouts/_default/index.json
Era-cell: I will need to read more about this. But in config.toml  file I shouldn't have an empty line at end right..
I think only in json file I should

jviotti: Not an empty line but a line ending. That's sounds like an issue with your editor configuration. Not sure what editor you use, but try to make sure it has https://editorconfig.org interoperability! That should pick up the `.editorconfig` files I have in most of the repos and always do the right thing!

jviotti (approved):

jviotti:
I fixed the line ending myself with a commit on top of your PR so we could merge. Other than that, looks and works great! Thanks!

layouts/_default/index.json
Era-cell: Then I think the formatter of my IDE did that. I get it a bit

PR #247, opened 2024-06-10
Fix `maxLength` example

dariocurr:
`This is a valid string` is 22 chars long


jviotti (approved):
Thank you so much for the heads up 🙏🏻 ❤️

PR #253, opened 2024-06-20
including an analogy to aid with schema resource terminology confusion

mathematikoi:


jviotti (approved):
Great analogy, thanks!

PR #258, opened 2024-07-05
Update if.markdown

IIIgnac:
'HOD_iD' in line 70 should be integer


jviotti (approved):
Great catch. Thanks a lot!

PR #269, opened 2024-11-06
fix(draft4): typos in descriptions

jeremyfiel:
couple typos fixed up

* updated attributes for some keywords

### DIFF
--- a/content/draft4/validation/additionalItems.markdown
+++ b/content/draft4/validation/additionalItems.markdown
@@ -1,8 +1,8 @@
 ---
 keyword: "additionalItems"
-signature: "Schema"
+signature: "Schema | Boolean"
 value: This keyword must be set to a valid JSON Schema
-summary: "If [`items`](/draft4/validation/items) is set to an array of schemas, validation succeeds if each element of the instance not covered by it validates against this schema."
+summary: "If [`items`](/draft4/validation/items) is set to an array of schemas, validation succeeds if each element of the instance not covered by it validates against this schema. If set to `false`, no additional items are allowed in the array instance."
 kind: [ "applicator" ]
 instance: [ "array" ]
 specification: "https://json-schema.org/draft-04/draft-fge-json-schema-validation-00#rfc.section.5.3.1"
--- a/content/draft4/validation/additionalProperties.markdown
+++ b/content/draft4/validation/additionalProperties.markdown
@@ -1,8 +1,8 @@
 ---
 keyword: "additionalProperties"
-signature: "Schema"
+signature: "Schema | Boolean"
 value: This keyword must be set to a valid JSON Schema
-summary: "Validation succeeds if the schema validates against each value not matched by other object applicators in this vocabulary."
+summary: "Validation succeeds if the schema validates against each value not matched by other object applicators in this vocabulary. If set to `false`, no additional properties are allowed in the instance."
 kind: [ "applicator" ]
 instance: [ "object" ]
 specification: "https://json-schema.org/draft-04/draft-fge-json-schema-validation-00#rfc.section.5.4.4"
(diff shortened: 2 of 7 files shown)


content/draft4/validation/additionalItems.markdown
```diff
@@ -1,8 +1,9 @@
 ---
 keyword: "additionalItems"
-signature: "Schema"
+signature: "Schema | Boolean"
```
jviotti, line 3: Actually, booleans _are_ schemas. JSON Schema defines a schema as either an object or a boolean, so not sure if this is required. If so, we would have to add such clarification everywhere we use the `Schema` type

content/draft4/validation/additionalItems.markdown
```diff
@@ -1,8 +1,9 @@
...
 value: This keyword must be set to a valid JSON Schema
-summary: "If [`items`](/draft4/validation/items) is set to an array of schemas, validation succeeds if each element of the instance not covered by it validates against this schema."
+summary: "If [`items`](/draft4/validation/items) is set to an array of schemas, validation succeeds if each element of the instance not covered by it validates against this schema.
+If set to `false`, no additional items are allowed in the array instance."
```
jviotti, line 6: I also think this is implicit from the fact the signature takes a schema. If you set it to the false schema that never validates, no additional items are permitted. 

Maybe better to discuss `false` in an example?

content/draft4/validation/additionalItems.markdown
jeremyfiel: Not sure that was true in earlier drafts. I'll look and try to clarify

jeremyfiel: Only from draft6 onwards was a boolean or schema defined explicitly. 

Prior to that it was considered an object. 

So I suppose my updates should be to later versions, rather than earlier  

What do you think? 

Draft6
https://json-schema.org/draft-06/draft-wright-json-schema-01#rfc.section.4.4

Draft4
https://json-schema.org/draft-04/draft-zyp-json-schema-04#rfc.section.3.2

Draft3
https://json-schema.org/draft-03/draft-zyp-json-schema-03.pdf  section 5

content/draft4/validation/additionalItems.markdown
jeremyfiel: with the change to Schema | Boolean, do you think it's ok to leave this description or do you prefer to move it to example still?

content/draft4/validation/additionalItems.markdown
jeremyfiel: i *think* it's only the ones I've defined.

content/draft4/validation/additionalItems.markdown
jviotti: Yep, let's keep the description! Though can you move it as part of the line above?

content/draft4/validation/additionalItems.markdown
jviotti: Let me do another read the spec just in case

jviotti: Also `properties` and `patternProperties`: https://json-schema.org/draft-04/draft-fge-json-schema-validation-00#rfc.section.5.4.4

content/draft4/validation/additionalProperties.markdown
```diff
@@ -1,8 +1,9 @@
 ---
 keyword: "additionalProperties"
-signature: "Schema"
+signature: "Schema | Boolean"
 value: This keyword must be set to a valid JSON Schema
-summary: "Validation succeeds if the schema validates against each value not matched by other object applicators in this vocabulary."
+summary: "Validation succeeds if the schema validates against each value not matched by other object applicators in this vocabulary.
+If set to `false`, no additional properties are allowed in the instance."
```
jviotti, line 6: I think this is breaking the build. Can you put it all in one line?

jeremyfiel:
fixed the build for these updates

jviotti (approved):

PR #288, opened 2025-03-23
Introduce a convenience shortcode to link to keywords

AgniveshChaubey:
I have extended the shortcode to accept two more arguments ('vocabulary' and 'dialect'), with 'dialect' being optional. @jviotti, the link you provided [here](https://github.com/sourcemeta/learnjsonschema.com/pull/244#discussion_r1624565786) for the `ref` shortcode seems to be broken, and with quick googling, I couldn't find a way to use one shortcode inside another.
BTW, I'm not sure if we need to use the `ref` shortcode here, as hugo by default returns a "Page Not Found" response when it cant find the page.


AgniveshChaubey:
Oops! I missed that. Actually I intentionally converted them to lowercase, basause the URL was not working with camelCase words.

I'll make a copy of the keyword before converting it to lowercase in order to preserve its case during render.

AgniveshChaubey:
In the latest changes, keyword casing has been preserved and `ref` shortcode have been used to catch broken links. Also, as per the suggestion, now, the `link` shortcode now renders vocabulary name inside parentheses
image

jviotti:
What do you think about mentioning the vocabulary? Does it feel OK to you, or is it too much?

AgniveshChaubey:
I'd say let's avoid mentioning vocabulary along with keywords. If we really wanna include it, one thing we can do is include the vocabulary name inside a tool tip.

jviotti:
Sounds good. Let's do that then

AgniveshChaubey:
The latest changes should show the vocabulary name inside a tootip. I wasnt able to test that because not sure why all of a sudden I started getting the following error:

> ERROR js.Build failed: Could not resolve "@popperjs/core"
WARN  found no layout file for "html" for kind "taxonomy": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.
Built in 1465 ms
Error: error building site: JSBUILD: failed to transform "main.js" (text/javascript): "C:\Users\agniv\Documents\learnjsonschema\assets\main.js:3:17": Could not resolve "fuse.js"
make: *** [Makefile:12: all] Error 1

I tried deleting node_modules and package-lock.json and then reinstalled them to make sure all the dependencies ware present, but the error still persists.

The changes should show the tooltip as expected. Just give a confirmation for that so that slowly I can start replacing all the keyword occurances with the `link` shortcode.

jviotti (approved):
Awesome. Looks great! Let's get this one merged and feel free to send more PRs incrementally updating things everywhere. Small PRs per keyword are probably better than a big one. I'm doing various updates to the site, so it will help avoiding potential conflicts.

PR #291, opened 2025-05-02
Fix/enum-keyword

jeremyfiel:
5.20. enum
 The value of this keyword MUST be an array. This array SHOULD have
 at least one element. Elements in the array SHOULD be unique.

Prior to draft5, the text reads:

5.5.1.1. Valid values
The value of this keyword MUST be an array. This array MUST have at least one element. Elements in the array MUST be unique.


content/2019-09/validation/enum.markdown
```diff
@@ -11,6 +11,8 @@ tests:
   - draft2019-09/enum.json
 index: -99998
 introduced_in: draft1
+changed_in:
+  - draft5
```
jviotti, line 15: Draft 5 was considered a minor yank revision to Draft 4. Can we say Draft 4 instead?

jviotti:
Interesting! Actually it's a bit more subtle. While the spec indeed says SHOULD in some cases, the meta-schemas often still enforce this:

- Draft 4: https://json-schema.org/draft-04/schema#. It has `minItems: 1`
- Draft 6: https://json-schema.org/draft-06/schema#. It has `minItems: 1`
- Draft 7: https://json-schema.org/draft-07/schema#. It has `minItems: 1`
- 2019-09: https://json-schema.org/draft/2019-09/meta/validation. Allows any number of items
- 2020-12: https://json-schema.org/draft/2020-12/meta/validation. Allows any number of items

Because many people will try to validate their schemas against the official meta-schemas, can we only drop the note for 2019-09 and 2020-12? 

And in that case, say it was changed in 2019-09?

jviotti:
Also there is a conflict on the PR!

content/2019-09/validation/enum.markdown
jviotti: Actually see my other comment. Seems like it was really relaxed only in 2019-09

jeremyfiel:
I also have a question on slack in the spec channel for the behavior of an empty enum.  Seems it should be unconstrained but it evaluates like a completely closed schema.

jviotti:
You could also bring this SHOULD for Draft 4, Draft 6, and Draft 7 to the team in Slack? Maybe it makes sense relaxing the meta-schemas accordingly if the specs only say SHOULD?

.devcontainer/devcontainer.json
jviotti, line 1: Interesting, but can we merge this as its own PR?

jeremyfiel: I dunno what this is.. something from the  codespace maybe   I'll update the pr to remove it

jeremyfiel:
> Also there is a conflict on the PR!

fixed up!

jviotti (approved):
Awesome thanks a lot!

PR #292, opened 2025-05-03
Chore/devcontainer-config

jeremyfiel:
This adds a codespace devcontainer configuration to make sure the Hugo Extended version is installed for all environments. 

I found using Win 10, the incorrect Hugo version is installed when using a GH Codespace, which doesn't allow for the use of the `make` command to build the docs locally without an error related to SASS

```
TOCSS: failed to transform "/main.scss"
```

This should provide more consistency for contributors. 

_ref: https://stackoverflow.com/a/55957784/8564731_


jviotti (approved):
I'm not sure how to try it out myself, but I trust it works for you! Thanks!

